PR #67, opened 2024-02-18
feat: Add magika docker file

Rhevin:
Add Docker file for Magika Python

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,16 @@ Magika is available as `magika` on PyPI:
 $ pip install magika
 ```
 
+### Running on Docker
+
+```
+git clone https://github.com/google/magika
+cd python
+docker build -t magika .
+docker run -it -v $path:/magika magika -r /magika
+
+Replace `$path` with target path
+```
 
 ### Usage
 
--- a/python/Dockerfile
+++ b/python/Dockerfile
@@ -1,0 +1,11 @@
+# syntax=docker/dockerfile:1
+
+ARG PYTHON_VERSION=3.11
+FROM python:${PYTHON_VERSION}-slim as base
+
+WORKDIR /magika
+
+RUN --mount=type=cache,target=/root/.cache/pip \
+    pip install magika
+
+ENTRYPOINT ["magika"]
--- a/python/.dockerignore
+++ b/python/.dockerignore
@@ -1,0 +1,33 @@
+# Include any files or directories that you don't want to be copied to your
+# container here (e.g., local build artifacts, temporary files, etc.).
+#
+# For more help, visit the .dockerignore file reference guide at
+# https://docs.docker.com/go/build-context-dockerignore/
+
+**/.DS_Store
+**/__pycache__
+**/.venv
+**/.classpath
+**/.dockerignore
+**/.env
+**/.git
+**/.gitignore
+**/.project
+**/.settings
+**/.toolstarget
+**/.vs
+**/.vscode
+**/*.*proj.user
+**/*.dbmdl
+**/*.jfm
+**/bin
+**/charts
+**/docker-compose*
+**/compose*
+**/Dockerfile*
+**/node_modules
+**/npm-debug.log
+**/obj
+**/secrets.dev.yaml
+**/values.dev.yaml
+LICENSE


python/requirements.txt
```diff
@@ -0,0 +1,14 @@
+click==8.1.7
+coloredlogs==15.0.1
+flatbuffers==23.5.26
+humanfriendly==10.0
+magika==0.5.0
```
gaby, line 5: Why is magika here? This will override line 45 of the Dockerfile

README.md
```diff
@@ -58,6 +58,14 @@ Magika is available as `magika` on PyPI:
 $ pip install magika
 ```
 
+### Running on Docker
+
+```
+git clone https://github.com/google/magika
+cd python
+docker build -t magika .
+docker run -it -v $(path to mount):/magika --rm magika magika ./$filename
```
gaby, line 67: Probably worth adding a workflow that publishes the image to ghcr.io

python/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+# into this layer.
+RUN --mount=type=cache,target=/root/.cache/pip \
+    --mount=type=bind,source=requirements.txt,target=requirements.txt \
+    python -m pip install -r requirements.txt
+
+# Switch to the non-privileged user to run the application.
+USER appuser
+
+# Copy the source code into the container.
+COPY . .
```
gaby, line 45: This should copy the python folder in, then install dependencies

gaby (changes requested):

python/requirements.txt
Rhevin: thanks you're right, i miss this using poetry
should be fixed

python/Dockerfile
Rhevin: thank you

README.md
Rhevin: Maybe separate pr?

gaby: Yeah, agree

CombatReq:
оу ### ### 
image


- [ ] @- > [_**бля******_](url)
- [ ]

invernizzi:
Thanks for this contribution.
This docker setup is valid, but it uses Magika from source. 


That's more bleeding edge that we'd want - people using magika in docker are either looking to try it out in a clean way, or they want to use it in a more "production" environment.
For that reason, I'd prefer the docker image to use published releases of the Python package.
Something like this should be enough for that purpose:
```Dockerfile
FROM python:${PYTHON_VERSION}-slim
pip install magika
ENTRYPOINT ["magika"]
```

gaby:
> Thanks for this contribution. This docker setup is valid, but it uses Magika from source.
> 
> That's more bleeding edge that we'd want - people using magika in docker are either looking to try it out in a clean way, or they want to use it in a more "production" environment. For that reason, I'd prefer the docker image to use published releases of the Python package. Something like this should be enough for that purpose:
> 
> ```dockerfile
> FROM python:${PYTHON_VERSION}-slim
> pip install magika
> ENTRYPOINT ["magika"]
> ```

^ Agree, probably also adding an ENV for argument for passing the magika version

Rhevin:
Thanks for the feedback
I will update this today

Rhevin:
updated

python/Dockerfile
```diff
@@ -0,0 +1,11 @@
+# syntax=docker/dockerfile:1
+
+ARG PYTHON_VERSION=3.11
+FROM python:${PYTHON_VERSION}-slim as base
+
+WORKDIR /magika
+
+RUN --mount=type=cache,target=/root/.cache/pip \
```
reyammer, line 8: Is the /root/.cache/pip location best-practice? /cc @invernizzi

reyammer (commented):
Thanks for the PR! Looks good to me, I just have one question before merging.

python/Dockerfile
reyammer: yep I understand the rationale, I'm just wondering whether the specific path is what's considered best-practice, vs. something like /var/cache/pip? is there a well-defined best practice to where to put the cache?

reyammer:
Alright, merging as is, but FYI I'll then move this into the repo root; now that we install magika with pip install, there is no reason to put it in python/. Thanks!

PR #76, opened 2024-02-19
Recommend pipx installation for CLI only

bswck:


python/README.md
```diff
@@ -5,10 +5,16 @@ Use Magika as a command line client or in your Python code!
...
 ## Installing Magika
+As a library:
 ```bash
 $ pip install magika
 ```
 
+As a [command-line tool](#using-magika-as-a-command-line-tool) only:
+```bash
+$ pipx install magika
+```
```
gaby, line 16: There's no real reason for this. Pip is fine

bswck: 👍🏻

reyammer: Not sure I get it: is there a strong reason for including pipx in the docs? pip does seem enough?

bswck: > Not sure I get it: is there a strong reason for including pipx in the docs? pip does seem enough?

What I can say is just [exactly what Python Packing Authority says](https://pipx.pypa.io/stable/#overview-what-is-pipx). it just so happens that pipx is a popular[[1]](https://github.com/cookiecutter/cookiecutter#installation)[[2]](https://python-poetry.org/docs/#installing-with-pipx)[[3]](https://tox.wiki/en/latest/installation.html#via-pipx)[[4]](https://github.com/wntrblm/nox#installation)[[5]](https://virtualenv.pypa.io/en/latest/installation.html#via-pipx) way of installing applications for CLI uses only, due to global, dedicated virtual environments for every app. I am happy to promote it whenever I can, as it's a modern approach to leveraging tools from the Python ecosystem. I believe `magika` is primarily useful as a CLI tool.

reyammer:
Thanks, reworded and merged.

PR #78, opened 2024-02-19
Add Expected extensions output mode to Magika CLI

nhonx:
Add `-e / --expected-exts` mode to Magika CLI, which output one or several expected file extensions of input file, in case the input extension is missed or incorrect

### DIFF
--- a/python/magika/cli/magika.py
+++ b/python/magika/cli/magika.py
@@ -86,6 +86,13 @@ Send any feedback to {CONTACT_EMAIL} or via GitHub issues.
     help="Compatibility mode: output is as close as possible to `file` and colors are disabled.",
 )
 @click.option(
+    "-e",
+    "--expected-exts",
+    "expected_extensions",
+    is_flag=True,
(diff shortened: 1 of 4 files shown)


reyammer:
Thank you for the PR! I'll need to give some thoughts about it, we have received a similar request and need to think how to integrate it for the long term. Leaving this open for now for visibility; will follow up later on. Thanks!

reyammer:
yes, indeed. We are already working on v2, and adding many more extensions / types. your examples are very spot on. And sorry for not following up on this, didn't have time to think about this and we are currently pushing for starting new training round. Will get to this once I have a moment, and I agree we should have a feature in this direction!

nhonx:
Hi @reyammer , should I close this? As I see we have new code adding to `main` branch and leading to conflict on this PR.

reyammer:
Finally following up on this, sorry for the very long delay! The new CLI written in Rust has a similar feature, thanks for proposing! See https://github.com/google/magika/blob/main/rust/cli/README.md for more details.

PR #79, opened 2024-02-19
Update README.md

abhin2002:
"announcment"       --> "announcement"
"Magika onlys use"  --> "Magika only uses"


reyammer:
Fixed in another PR, but thank you anyways!

PR #80, opened 2024-02-19
feat: initialise and load class in 1 call (npm/js)

TheMythologist:
Continuation of #52 

It is discouraged from using async functions in constructor methods in JavaScript. Instead, I implemented an alternative way to initialise the `Magika` class in 1 call.

Now there are 2 ways (both will achieve the same results). This way, we have a more efficient way to initialise the class, but we also preserve the original way for backwards compatibility.
```js
const magika = new Magika();
await magika.load();
```

or

```js
const magika = await Magika.create();
```

Both `load` and `create` function take in an optional parameter with a default parameter.

### DIFF
--- a/js/magika.js
+++ b/js/magika.js
@@ -78,6 +78,12 @@ export class Magika {
     await Promise.all([this.config.load(configURL), this.model.load(modelURL)]);
   }
 
+  static async create(options = {}) {
+    const magika = new Magika();
+    await magika.load(options);
+    return magika;
+  }
+
   /** Identifies the content type of a byte stream.
    *
    * @param {*} fileBytes  a Buffer object (a fixed-length sequence of bytes)


reyammer:
Thank you! LGTM, but I'm not a js expert. Adding @invernizzi as reviewer, he will follow up.

In the meantime: I've reorganized the docs, we are now collecting them in `./docs/*.md`. So, `js/DOCS.md` moved to `docs/js.md`; this may need to be cleaned up in the PR. I've also updated the gh action so that the automatic script patches the right file. Sorry for this annoyance.

TheMythologist:
Apologies for the closing and opening of PR, but the merge conflict should be fixed now

TheMythologist:
We may also need to update the documentation, and point people to use the new way to initialise Magika via `Magika.create()`

TheMythologist:
Hi, any update regarding this?

invernizzi:
Thanks for your contribution! Apologies for the delay, there's lots going on with Magika these days

invernizzi (commented):
+1

PR #148, opened 2024-02-20
deps: add alternative numpy version for py3.12

### LINKED ISSUE #43: Add support for 3.12 python
Is there any reason that prevents it from running on 3.12 python?

BTW amazing job, great library!

TheMythologist:
Resolves #43 

> Reference: https://python-poetry.org/docs/dependency-specification/#multiple-constraints-dependencies

I noticed in [your `support-python-3.12` branch ](https://github.com/google/magika/tree/support-python-3.12) that you also bumped the version of the package `onnxruntime`. Do you know if that is necessary?

### DIFF
--- a/python/poetry.lock
+++ b/python/poetry.lock
@@ -341,6 +341,51 @@ files = [
 ]
 
 [[package]]
+name = "numpy"
+version = "1.26.4"
+description = "Fundamental package for array computing in Python"
+optional = false
+python-versions = ">=3.9"
+files = [
+    {file = "numpy-1.26.4-cp310-cp310-macosx_10_9_x86_64.whl", hash = "sha256:9ff0f4f29c51e2803569d7a51c2304de5554655a60c5d776e35b4a41413830d0"},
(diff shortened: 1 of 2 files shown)


reyammer:
Thanks for the PR! Super helpful. And yes, I needed to bump onnxruntime version as well (the version I had seemed to throw some errors when tried to install on py 3.12). But not sure whether we could specify an earlier version or whether it must be this one.

But since we would specify these only for py >=3.12, I guess it's not a big deal?

And if it's not a big deal, I'm wondering whether we should update the versions for all our deps...

I'm trying to think whether we could introduce future problems by forcefully bump the required version of these packages, or whether it's really not a problem...

gaby:
@reyammer I covered the changes from this PR in https://github.com/google/magika/pull/174 Upgraded both Numpy and onnxruntime to be able to support py3.12

reyammer:
@gaby this PR seems better, as it supports BOTH py 3.8 and py 3.12! For the py 3.12 support I'm taking this one, but yours also has good stuff on upgrading the other packages. I hope it doesn't create too many problems :-) I'll follow up with more comments in #174

PR #157, opened 2024-02-20
Removed MIME Type Column from supported  list

devilkadabra69:
new colums are ->Index, Content Type Label ,Description

### DIFF
--- a/docs/supported-content-types-list.md
+++ b/docs/supported-content-types-list.md
@@ -1,120 +1,120 @@
 # List of currently supported content types
 
-| Index   |      Content Type Label      |  MIME Type | Description |
-|----------|:-------------:|------:|------:|
-| 1 | ai | application/pdf | Adobe Illustrator Artwork |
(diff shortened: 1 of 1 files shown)


nhonx:
We also need to remove the Column from here as well:
https://github.com/google/magika/blob/a9d3904fbe1523c740ecb20ebf0e742e4acf5892/python/magika/cli/magika.py#L432
It's shown when user use `magika --list-output-content-types`

reyammer:
Thanks for the PR folks! The issue I've created was more of a "think about it" rather than "let's do it for sure" :-) but I still think it's a good idea. Maybe we should add a short discussion at the beginning or end of the table to justify this / point whoever sees the table to the discussion on output format. I need to think a bit about this...

devilkadabra69:
i guess i missed headers part ! i'll fix it quickly

reyammer:
Thanks!

PR #164, opened 2024-02-20
Updated-removal_mime_type Column

devilkadabra69:


reyammer (changes requested):
we need to remove line 438 as well. Can you take care of this and then test it? I would expect the current PR to crash / show bad output?

devilkadabra69:
can you just clearify :- other than line 438 where can it all be present because i was going through mime_type variable is present all over the code can you explain a bit

reyammer:
I think it's OK to delete only that line. we don't want to print the mime type if we remove the mime type header from the column.

devilkadabra69:
can you check the commit i have applied the necessary changes in the line 438

reyammer:
Hello, sorry I've just merged #250 and I thought it was your initial PR. Thanks anyways and sorry!

devilkadabra69:
no issues
its ok

Issue #47, opened 2024-02-16
Help needed: tests on other platforms (Windows, MAC OS, non-Ubuntu distros, etc.)


reyammer:

gaby:
Running the tests for Magika on a Windows platform with py3.11:

```console
============================= test session starts =============================
platform win32 -- Python 3.11.8, pytest-7.4.0, pluggy-1.2.0
rootdir: D:\a\magika\magika\python
configfile: pytest.ini
collected 41 items / 2 deselected / 39 selected

tests/test_magika_python_cli.py::test_magika_python_cli_with_one_test_file FAILED [  2%]
tests/test_magika_python_cli.py::test_magika_python_cli_with_very_small_test_files FAILED [  5%]
tests/test_magika_python_cli.py::test_magika_cli_with_small_test_files FAILED [  7%]
tests/test_magika_python_cli.py::test_magika_cli_with_empty_file FAILED  [ 10%]
tests/test_magika_python_cli.py::test_magika_cli_with_directories FAILED [ [12](https://github.com/gaby/magika/actions/runs/7982383780/job/21796809351#step:6:13)%]
tests/test_magika_python_cli.py::test_magika_cli_with_symlinks FAILED    [ [15](https://github.com/gaby/magika/actions/runs/7982383780/job/21796809351#step:6:16)%]
tests/test_magika_python_cli.py::test_magika_cli_with_files_with_permission_errors FAILED [ [17](https://github.com/gaby/magika/actions/runs/7982383780/job/21796809351#step:6:18)%]
tests/test_magika_python_cli.py::test_magika_cli_with_basic_test_files FAILED [ [20](https://github.com/gaby/magika/actions/runs/7982383780/job/21796809351#step:6:21)%]
tests/test_magika_python_cli.py::test_magika_cli_with_mitra_test_files FAILED [ [23](https://github.com/gaby/magika/actions/runs/7982383780/job/21796809351#step:6:24)%]
tests/test_magika_python_cli.py::test_magika_cli_with_basic_test_files_and_json_output FAILED [ [25](https://github.com/gaby/magika/actions/runs/7982383780/job/21796809351#step:6:26)%]
tests/test_magika_python_cli.py::test_magika_cli_with_basic_test_files_and_jsonl_output FAILED [ [28](https://github.com/gaby/magika/actions/runs/7982383780/job/21796809351#step:6:29)%]
[...]

Issue #88, opened 2024-02-19
Test support for Mac OS


reyammer:

gaby:
@reyammer This should be doable using githuh workflows. I can send a Pull Request

reyammer:
Thanks! Is there a way to test the actual magika package that people install from pip, rather than just running the tests in a macos environment? my current gh workflow does the latter, but we may want to improve.

invernizzi:
Much appreciated, @gaby! 
Ideally, we'd like to build some decent CI for all sort of platforms. So, bonus points if your PR can be used to start something like that.

reyammer:
IIRC you can specify something like this in gh action: `os: [ubuntu-22.04, macos-latest, windows-latest]`. Is this the direction, @gaby?

But independently from this, the current CI does "poetry install + pytests tests"... which is a good starting point, but not THE tests I would also like to have... ideally we should test for "what people install with pip install magika". Created #158 to track this additional aspect.

gaby:
@reyammer Correct, that will run the tests in that enviroment.

I was thinking that you could have a workflow that does the pip install and runs against the `tests_data/` directory to verify it can run on that platform

ehjmx:
I have Python 3.12 installed via MacPorts on macOS 14.3.1. on a MacBookPro 13inch M2

When I try to install magika with pip312 I get the following error message:
```
pip-3.12 install magika
Defaulting to user installation because normal site-packages is not writeable
ERROR: Ignored the following yanked versions: 0.1.0
ERROR: Ignored the following versions that require a different python version: 0.5.0 Requires-Python >=3.8,<3.12
ERROR: Could not find a version that satisfies the requirement magika (from versions: none)
ERROR: No matching distribution found for magika
```
I haven't tried the pip3 which comes with the latest XCode. It might work as python3 --version gives me 3.9.6

gaby:
@ehjmx Probably needs to be installed in a venv. The dependencies are conflicting with the ones from the OS

gaby:
I'm updating the workflows to run on multiple platforms/python version. I can confirm that the tests only pass on `linux` and `macos`. When using `windows` a significant amount of tests are failing.

PR #165, opened 2024-02-20
pytest support for windows and mac OS (platform-independent), text (.txt) sample file added to basic test data

abhin2002:
#47
#88


gaby:
@abhin2002  This doesnt fix #47 #88 The tests need to be run on those platforms. This is just running tests on the default platform

PR #174, opened 2024-02-21
Support for py3.12, and MacOS testing

### LINKED ISSUE #88: Test support for Mac OS

gaby:
- Add support for dependabot to track `python`, `nodejs`, `github-actions` dependencies
- Upgraded `pytest` to `8.0.x`
- Support for Python `3.9` required the removal of the `|` operator.
- Added support for running unit-tests in `macos-latest`
- Fixed the poetry lock file that was broken from #148 
- Fixed versions of Numpy based on Python version.
- Fixed versions of `ipython` based on Python version.

Note: Unit-test fail with `windows-latest`, that should be tracked in a separated PR.

fixes #88


gaby:
Tried cleaning my commits but made it worst. At least it's not +15 commits

reyammer:
Hello @gaby, thanks for the PR and sorry for the conflicts mess. It's tricky to review and merge this one as it touches on too many things. It would be really helpful if we could split this PR in different ones, this one touches py 3.12 stuff, mac os support, deps upgrade, and various other fixes. All this is valuable, but if I merge it as is and something breaks, I wouldn't know where to start debugging :)

gaby:
The lock file was not updated in #148 which is causing the unit-tests to fail. Working on doing that now

gaby:
@reyammer This should be good to review now. The lock file was broken from #148, fixed that. Also Numpy 1.26 can be used since 3.9, so fixed that too. 

You can see the unit-test results here: https://github.com/gaby/magika/actions/runs/7990847889

python/pyproject.toml
```diff
@@ -25,7 +22,7 @@ python-dotenv = "^1.0.1"
 magika = "magika.cli.magika:main"
 
 [tool.poetry.group.dev.dependencies]
-pytest = "^7.4.0"
+pytest = ">=7.4,<9.0"
```
reyammer, line 25: what is the difference between ^7.4 and >=7.4?

python/poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 1.7.1 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.5.1 and should not be changed by hand.
```
reyammer, line 1: This seems generated by an old version of poetry. Seems OK anyways?

reyammer (commented):
overall everything looks good (just left a couple of questions). will merge tomorrow morning so that I'm around if something goes south. Thanks!

python/poetry.lock
gaby: Fixed, I had 2 poetry installs on my VM

python/pyproject.toml
gaby: This actually needed to be fixed, pytest is now v8.0.1, while doing this also fixed `ipython` which versions change based on your Python version.

gaby:
> overall everything looks good (just left a couple of questions). will merge tomorrow morning so that I'm around if something goes south. Thanks!

Sounds like a good idea, turns out during my squash I forgot some commits. All dependencies are updated now. Fixed the version of `ipython` too, it now gets a different version based on your python version.

Test are passing:
- mypy: https://github.com/gaby/magika/actions/runs/7999193931
- unit-tests: https://github.com/gaby/magika/actions/runs/7999193934

reyammer:
Merged, thank you!

Just added a quick fix to the ipython dep specification, as it seems there was no valid candidate for python 3.12? The tests don't cover that because it's a dev dependency, would be nice to extend the tests to cover those as well.

gaby:
@reyammer Let me see what I can find, there's probably a way to tests that in a separate workflow.

PR #175, opened 2024-02-21
text (.txt) sample file added to basic test data

abhin2002:


reyammer:
Thanks for the test case! But I see in the commit history something unrelated (windows testing, etc.). can you send a clean PR with just this new test case?

And about windows testing: we do need help, so feel free to open an issue to start discussion in case you found problems. Thanks!

PR #253, opened 2024-02-24
Refactor Python GitHub Actions

gaby:
- Run `ruff check` and `ruff format check` during CI.
- Merge `mypy` CI into the same pytest CI this way dependencies dont have to be installed twice
- Bump ruff to `0.2.2`
- Add `**/.ruff_cache/` to gitignore
- Renamed workflow to "Python", that way when more languages are added the workflow names are not super long.
	- For example for e2e it will say "Python / e2e-testing"


gaby:
@reyammer Small refactoring of Python CI to reduce duplication.

PR #254, opened 2024-02-24
Add e2e testing of Python package using GitHub Actions

### LINKED ISSUE #158: Improve automated tests to cover what people install with `pip install magika`, rather than `poetry` environment.
The current CI does the equivalent of `poetry install && pytest tests`. This is great to test the current code on github.

However, we currently have no tests for "what people actually install" with `pip install magika`. The assumption is that the poetry environment / installation is the same as what people get with `pip install magika`. So far this assumption has hold true (we have no reported issues on this), but I would feel better if we would start automatically testing more closely what's actually installed by users. This feels even more important for platforms we are not so familiar with, e.g., Windows and Mac OS.

gaby:
Automates the process of building `magika` python wheel, installing the wheel and running `magika` against the files in `tests_data/`. If unable it will fail the CI job.

The test is run against py38 -> py3.12 using both Linux, MacOS, and Windows.

image

Fixes #158

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -1,4 +1,4 @@
-name: E2E Testing
+name: Python
 on:
     push:
       branches:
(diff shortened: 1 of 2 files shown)


gaby:
@reyammer Is this what you envisioned with #158 ?

PR #255, opened 2024-02-25
Add gh action to label new issues

### LINKED ISSUE #155: Setup a github action (or similar) to apply the "needs triage" label to every new external github issue
This helps us tracking which issue still needs the first pass.

Checking online it seems there is no way to apply a label by default to every template, and it seems the way to go is to add a github action?

Ideally, the label would not be applied if one of the current maintainers is opening it. But that's not so important...

Mohit-Gaur:
Resolves #155 

Setup a GitHub action that adds the "needs triage" label to new issues.

Tested it by creating a sample issue in my fork [here](https://github.com/Mohit-Gaur/magika/issues/5) and its working fine.


Mohit-Gaur:
**Note** - This action will currently label all new issues regardless of whether it was created by a project maintainer or external user.

There are two ways to determine if an issue creator is a maintainer:

1. Use the GitHub API to check if user is a collaborator on the repo.
2. Use a CODEOWNERS file to maintain a list of project maintainers. More info [here](https://github.blog/2017-07-06-introducing-code-owners/).

I am leaning towards 2. But would like to hear your thoughts. Thanks!

.github/workflows/new-issue-labeler.yml
```diff
@@ -0,0 +1,20 @@
...
+on:
+  issues:
+    types: [opened]
+
+jobs:
+  label:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Label new issues
+      uses: actions/github-script@v7
```
gaby, line 11: There is an official action for this: https://github.com/github/issue-labeler

The workflow can be made optional by using the context variables.

gaby (changes requested):

.github/workflows/new-issue-labeler.yml
Mohit-Gaur: Thanks for sharing. Updated the PR!

While we can use context variables here, I'm not sure if there's a way to get the role of an issue author using that.

reyammer:
Thanks!

PR #274, opened 2024-03-01
feat: move js lib to typescript, add tests and add support for file streams

anzerr:
Hello, I needed support for file streams so I could run this on larger files. I wanted to replaced a lib in a older project with this but it needed CJS support. I also noticed that the JS lib didn't have tests yet. 

So I made a few changes:
- Added typescript to leverage the typing from TensorFlow.js
- Added jest with a few basic tests
- Added identify support from a read steams
- Fixed the middle chunk offset being wrong
- Added the ability to load the model from a local file
- Added hybrid support for CJS and ESM


ebursztein:
Thanks for the PR ! Those improvements seems very useful.  Assigning to @invernizzi .

js/src/model.ts
```diff
@@ -0,0 +1,80 @@
+import * as tf from '@tensorflow/tfjs';
+import * as tfn from '@tensorflow/tfjs-node';
```
invernizzi, line 2: This import breaks the browser package, as `tfjs-node` is not available there. 
Since `tfn` is only used once,  can you import `tfjs-node` dynamically instead? 
Removing this line and adding something like `const tfn = await import('@tensorflow/tfjs-node')` on line 48  (+ some error checking) should fix this.

To test that it works properly, run:
```
cd js
npm run build
npm link
cd ../website
npm install
npm link 'magika'
npm run dev
```

Right now, you'd get this output:

```

  VITE v5.1.4  ready in 348 ms

  ➜  Local:   http://localhost:5173/magika/
  ➜  Network: use --host to expose
  ➜  press h + enter to show help
✘ [ERROR] No loader is configured for ".html" files: ../js/node_modules/@mapbox/node-pre-gyp/lib/util/nw-pre-gyp/index.html

    ../js/node_modules/@mapbox/node-pre-gyp/lib/node-pre-gyp.js:86:21:
      86 │       return require('./' + command)(self, argvx, callback);
         ╵                      ~~~~~~~~~~~~~~

✘ [ERROR] Could not resolve "mock-aws-s3"

    ../js/node_modules/@mapbox/node-pre-gyp/lib/util/s3_setup.js:43:28:
      43 │     const AWSMock = require('mock-aws-s3');
      [...]
 ```

invernizzi:
Thank you for the PR! I've tested it and it works great for Node. I highly appreciate the tests.
The browser version has a conflict at the moment, but I'm sure it can be fixed (if you don't have time for that, please let me know).

anzerr:
> Thank you for the PR! I've tested it and it works great for Node. I highly appreciate the tests. The browser version has a conflict at the moment, but I'm sure it can be fixed (if you don't have time for that, please let me know).

Thanks for the feed back, I updated the pr with the changes. I completely forgot about the impact the import would have. I updated the tests to cover the dynamic loading of `tfjs-node` to try and catch a regression.

invernizzi:
Spectacular, thanks for the quick update!

invernizzi (commented):
merged

PR #298, opened 2024-03-11
Add check-text flag to return whether file is text or binary format

nhonx:
Use flag `--check-text` to return if file is Text format.


reyammer:
Closing for now, let's discuss in #303.

Issue #221, opened 2024-02-22
add coverage analysis for tests of the python package


reyammer:

ps2181:
I have created a .ini for running the coverage analysis and have gotten the report as a html file and its generated locally. Could you please suggest me what to do next?

reyammer:
Hello, can you create a new PR (referencing this issue) and we take it from there? The overall idea is to start monitoring the code coverage of our tests, add it to one of the existing workflows, and, later on, display some coverage badge in our README. I'm not sure about the details, but there seems to be many posts describing the process (e.g., https://nedbatchelder.com/blog/202209/making_a_coverage_badge.html). Thanks!

ps2181:
Hello,  I have created a pull request addressing the code coverage monitoring for our tests as referenced in Issue #221.

In the PR, I have implemented the necessary changes to start monitoring code coverage . Please let me know if this aligns with your expectations or if there are any additional changes or details needed.

below are the screenshots of the code coverages 
1. using pytest
![Image](https://github.com/user-attachments/assets/a01fa59f-5e00-4b1a-8481-113d330db405)

2. using tox
![Image](https://github.com/user-attachments/assets/c6b842de-f216-4dc4-8dc3-db80e38777a2)

PR #316, opened 2024-03-22
code coverage added in workflows

abhin2002:
Closes issue #221

### DIFF
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -54,3 +54,39 @@ jobs:
       - name: Run pytest
         working-directory: python
         run: poetry run pytest tests -m "not slow"
+
+      - name: Code Coverage Report
+
+        uses: irongut/CodeCoverageSummary@v1.3.0
+
+        with:
+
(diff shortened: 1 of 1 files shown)


.github/workflows/python-test.yml
```diff
@@ -54,3 +54,39 @@ jobs:
...
+
+          hide_complexity: true
+
+          indicators: true
+
+          output: both
+
+          thresholds: '60 80'
+
+    - name: Add Coverage PR Comment
```
invernizzi, line 82: Thank you for your contribution!
The indentation of this line breaks the parsing of the file. Could you fix it?

invernizzi (changes requested):

.github/workflows/python-test.yml
abhin2002: I hope indentation fixed

invernizzi:
Thanks! It seems that the check wasn't successful because no code coverage data is actually being produced. We need to modify how pytest is called so that it generates that data.

invernizzi:
@dependabot rebase

PR #328, opened 2024-03-26
Switch Rust dependency from onnxruntime to ort

ia0:
Also add Rust tests to CI.


.github/workflows/rust-test.yml
ia0, line 1: @reyammer Can you take a look at this file? I copied from `python-test.yml`.

invernizzi (approved):

PR #406, opened 2024-04-12
Split Rust code between CLI and library

ia0:
Also use the latest release candidate of the ORT library, which provides async inference. The CLI now creates only one session and shares it between the different inference tasks.

This also implements the features extraction reference tests.


rust/gen/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+        let mime = unwrap(config.mime_type, "mime_type", &short)?;
+        let group = unwrap(config.group, "group", &short)?;
+        let magic = unwrap(config.magic, "magic", &short)?;
+        let long = unwrap(config.description, "description", &short)?;
+        let text = config.tags.iter().any(|x| *x == "text");
+        let label = Label { short, long, magic, group, mime, exts, text };
+        labels.entry(target).or_default().push(label);
+    }
+    let labels = labels;
+    let mut output = File::create("../lib/src/label.rs")?;
```
invernizzi, line 47: Could you add a comment explaining why dynamically generating this file is necessary (or add it to the README.md of this directory)?

rust/lib/src/label.rs
invernizzi, line 1: please add a line mentioning that this is generated by the gen crate. It should help code wanderers to figure things out.

Issue #93, opened 2024-02-19
Make application / bindings for Rust


reyammer:
When using magika as a python command line tool, the biggest overhead comes from starting the python interpreter itself and loading all the libraries.

Onnx's model loading itself is relatively fast (~30ms), the content type inference is ~5ms. But the overall python CLI takes about ~300ms.

Having a client written in Rust may make the CLI 8x/9x/10x faster.

This is not a concern for large scale automated pipelines (as they would boostrap the library and load the model only once), but it is annoying for one-off CLI use cases.

If this works out, then we could make this new rust client the one installed via `pip install magika` (together with the python API).

We may also want to have a proper way to make magika available to rust applications.

Note for external contributors: this may become the future de-facto magika client, so its design will need extra care. We are reaching out internally as well, so please let us know if you are starting working on this so that we can coordinate better and/or avoid duplicated work. Thank you!

reyammer:
Update: We have some POC working and someone working on it. Placeholder in #180.

maekawatoshiki:
I'm aware that magika written in Rust is entirely still in progress, but using `onnxruntime` crate (last updated 2yrs ago) in it does not seem reasonable. How about using [ort](https://github.com/pykeio/ort) crate?

reyammer:
Thanks for raising this! @ia0 thoughts?

ia0:
The PR switching from `onnxruntime` to `ort` is:
- #328

cr-itay-baranes:
@ia0 
What should be the Rust impl equivalent of the python Result.output.ct_label?
It seems that only the deep learning model result is reflected in the Rust impl, right ?

ia0:
> It seems that only the deep learning model result is reflected in the Rust impl, right ?

Yes that's correct. At the moment the Rust library only provides access to the deep learning model. It is planned (soon) to make it closer to the Python library (including computing ct_label without deep learning for small files).

> What should be the Rust impl equivalent of the python Result.output.ct_label?

This should be `magika::Label`. It already contains the `Txt` and `Unknown` labels, but they are not yet used when deep learning is bypassed (because that bypass is not yet implemented, see above).

ia0:
[...]

PR #488, opened 2024-06-03
Implement get_result_of_few_bytes in Rust

ia0:
This should address https://github.com/google/magika/issues/93#issuecomment-2137550297.

Changes in this PR:
- [lib] Add a `FeaturesOrOutput` type for the result of features extraction, which may now directly produce an output.
- [lib] Add two helpers to identify a file directly from its input without having to compute features first.
- [cli] Support reading from standard input.


ia0:
Closing in favor of a single big PR: #536

PR #490, opened 2024-06-04
Implement same colored output in Rust CLI as Python

ia0:
In particular, show the group instead of the score by default.


ia0:
Closing in favor of a single big PR: #536

PR #554, opened 2024-07-05
New release for the python module + new model

reyammer:
This is still an heavy WIP.


reyammer:
Note: two tests are currently not passing: one is about integration with rust (which we'll fix with @ia0), one is about the end-to-end tester, which expects a "magika" executable in the path (this is now broken because we have deprecated the python CLI, and we still need to package the rust CLI in the python package).

invernizzi (approved):

python/pyproject.toml
```diff
@@ -1,9 +1,10 @@
 [tool.poetry]
 name = "magika"
-version = "0.5.2-dev"
-description = "A tool to determine the content type of a file with deep-learning"
-authors = ["Yanick Fratantonio <yanickf@google.com>"]
+version = "0.6.0-dev"
+description = "A tool to determine the content type of a file with deep learning"
+authors = ["Magika Developers <magika-dev@google.com>"]
```
ia0, line 5: Is this something we want to do in Rust too?

ia0 (approved):

python/pyproject.toml
reyammer: I would? That's so that everyone is in the loop, having myself as the only POC was not intended for python :-)

ia0: Sounds good, I'll do that.

PR #564, opened 2024-07-17
Use Maturin to ship the Rust CLI in the Python package

ia0:


reyammer (approved):

invernizzi (dismissed):

python/pyproject.toml
```diff
@@ -28,8 +28,15 @@ ruff = ">=0.2.2,<0.6.0"
...
-requires = ["poetry-core"]
-build-backend = "poetry.core.masonry.api"
+requires = ["maturin>=1.7,<2.0"]
+build-backend = "maturin"
+
+[tool.maturin]
+bindings = "bin"
+manifest-path = "../rust/cli/Cargo.toml"
+locked = true
+compatibility = "linux"
```
reyammer, line 38:
```suggestion
compatibility = "linux"
module-name = "magika"
python-source = "python"
```

ia0:
@reyammer Do we want to also do an `sdist` wheel?

.github/workflows/maturin.yml
```diff
@@ -0,0 +1,42 @@
+# This file is was initially generated by maturin generate-ci github, then heavily simplified.
+
+name: Maturin
+
+on:
+  push:
+    branches:
+      - 'main'
+    tags:
+      - '*' # TODO: which tags matter?
```
ia0, line 10: How do we plan to tag releases? So that we only generate the wheels artifacts on release tags.

.github/workflows/maturin.yml
```diff
@@ -0,0 +1,42 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Build wheels
+        uses: PyO3/maturin-action@v1
+        with:
+          container: off # The build fails in containers.
+          working-directory: python
+          args: --release --out=../dist
+      - name: Upload wheels
+        uses: actions/upload-artifact@v4
```
ia0, line 39: TODO(ia0): Add an if-condition to only run on the specified tags.

.github/workflows/maturin.yml
reyammer: this is the pattern we have been using for python and js: `python-v0.5.0`, so something like `rust-v*` should work?

.github/workflows/maturin.yml
```diff
@@ -0,0 +1,55 @@
...
+            args: --release --out=../dist
+          - name: windows
+            runner: windows-latest
+            command: build
+            args: --release --out=../dist
+          - name: macos
+            runner: macos-latest
+            command: build
+            args: --release --out=../dist
+          # TODO(https://github.com/PyO3/maturin/issues/2154): Re-enable when fixed.
```
ia0, line 36: We can only provide an sdist if we don't ship the README.md.

reyammer (approved):

PR #584, opened 2024-07-22
Release Rust crates

ia0:


Menguozi:
Can you please release magika crate for rust soon? We have a project on container image that needs to experience this dependency library. Thank you for your commitment to open source! _(:з」∠)_

reyammer (approved):

reyammer:
@Menguozi: we are getting there! We should be able to release the first -rc release by next week, and we'd love to get some feedback :-)

ia0:
This commit has been published to crates.io.

Because it is a pre-release, one needs to explicitly ask for this version (cargo doesn't select pre-releases):
- To install the CLI: `cargo install --locked magika-cli@0.1.0-rc.0`
- To use the library: `magika = "0.1.0-rc.0"`

Note that there should be a 0.1.0-rc.1 coming up by next week (which should be a non-breaking change because only the model would change).

PR #632, opened 2024-08-23
tests: add current/past missdetections and improve python testing

reyammer:


ia0 (approved):
LGTM. If I understand correctly, this is doing the same label test based on the directory name for the new `previous_missdetections` directory as it is currently doing for the `basic` and `mitra` directories. I can update the equivalent Rust tests in this PR as an additional commit or in a follow-up PR. As you want.

reyammer:
Correct; it's extending the tests to problematic files that the new version deals with correctly. if it's easy to update for rust as well, sure let's do it. Thanks!

PR #638, opened 2024-08-26
KB: update content types KB

reyammer:


reyammer:
@ia0 I've updated the KB (mostly metadata), and it seems some rust tests "expected output" needs to be regenerated

ia0:
> @ia0 I've updated the KB (mostly metadata), and it seems some rust tests "expected output" needs to be regenerated

You just need to follow the error message. I'm copying it here: 

    Todo: Execute ./sync.sh from the rust directory

reyammer:
Fixed mime types typos and run rust sync. Should be good to go.

However: in the meantime we added a new content type to the KB ("AU"), it's not used for the model at the moment, but it does appear in the KB... but the rust list of content types doesn't show any diff related to that. My assumption is that that's the case because AU is not used by the model => rust doesn't care. Is that correct?

rust/lib/src/content.rs
```diff
@@ -1651,9 +1651,9 @@ pub(crate) static TIFF: TypeInfo = TypeInfo {
 
 pub(crate) static TOML: TypeInfo = TypeInfo {
     label: "toml",
-    mime_type: "text/plain",
+    mime_type: "application/tom",
```
ia0, line 1654: typo?

```suggestion
    mime_type: "application/toml",
```

ia0 (approved):
I don't guarantee I didn't miss any obvious mistake, since I'm anyway not familiar with even half those content types.

ia0:
> AU is not used by the model => rust doesn't care. Is that correct?

Yes. We only embed content types that are used (either historically by previous models, or currently by the embedded model).

rust/lib/src/content.rs
reyammer: good catch, fixed

ia0 (approved):

PR #639, opened 2024-08-26
update models name, remove draft_label

reyammer:
- clean up list of currently (publicly) supported models
- remove draft_ prefix, and add _X as minor version number (`draft_standard_v2` => `standard_v2_0`)


reyammer:
@ia0 FYI

python/magika/magika.py
```diff
@@ -40,7 +40,7 @@
     StatusOr,
 )
 
-DEFAULT_MODEL_NAME = "draft_standard_v2"
+DEFAULT_MODEL_NAME = "standard_v2_0"
```
ia0, line 43: To replicate this change in Rust, you need to update the `rust/gen/model` symbolic link to point to the new directory. You'll need to run the sync script, similarly to the other PR because the model name changed.

reyammer:
Done, should be good to go.

ia0 (approved):
LGTM for rust directory only

Issue #640, opened 2024-08-27
Clean up magika assets managment


reyammer:
Currently, models and other assets such as KB lives in the python/magika/models dir, with the rust (and js?) codebases pulling things from there. This is not ideal as we may want to decouple the various codebases (and we may want to upgrade one model for python at a different time than for rust). Also, we often work with .onnx format, but for some other stuff we may want the .h5 or the tfjs version... I could store them all in python/magika/models, but they don't belong there... at the moment, whatever is in python/magika/ gets packaged with the pip package, which is not good.

My proposal: I update my code from the training side of things / internal repo to export assets (models + content types KB + whatever else) somewhere like <root_github_repo_dir>/assets/*. Then, the various codebases (python, rust, go, js, etc.) can pick from that folder whatever they want/need (in an automated or manual fashion).

Thoughts?

CC @invernizzi @ia0

ia0:
I also think this is a better design. So all green on my side to proceed. Once you do the base layout, I can update the Rust side of things. Just ping me on the PR.

reyammer:
#643 stores assets in `assets/`, which is the first step. Then, we need to clean up the `assets_generation/` script.

reyammer:
Closing this one as we actually cleaned up `/assets`. Opened #684 to track what we need to clean about `/assets_generation`.

PR #643, opened 2024-08-28
store KB and models to assets dir

reyammer:
Store all assets in `assets/`. The various codebases (python/, rust/, js/, ...) can then pick up whatever asset they need at their own time (in a manual or automatic manner, via gh actions). This way, we start to decouple the various codebases from the python one.


reyammer:
Also added `metadata.json` to the model asset dir. This is not needed to run the model and it's not part of the config; it's just metadata to keep track how the model was created / where it comes from. For now, it just contains the hash of the namespace and model_name, as tracked by our training pipeline.

reyammer:
BTW #2: the idea is that the assets for each model contain 1) all the formats we may need for the various codebases (h5/keras, onnx, tfjs), 2) all the info we need to generate documentation about a given model, e.g., which content types are supported.

reyammer:
This is related to #640.

assets/models/standard_v2/config.min.json
```diff
@@ -0,0 +1 @@
...95, "pickle": 0.95, "png": 0.95, "po": 0.95, "postscript": 0.95, "powershell": 0.6, "ppt": 0.5, "pptx": 0.95, "prolog": 0.95, "proteindb": 0.95, "proto": 0.95, "psd": 0.95, "python": 0.95, "pythonbytecode": 0.95, "qt": 0.95, "r": 0.95, "rar": 0.95, "rdf": 0.95, "rpm": 0.95, "rst": 0.95, "rtf": 0.95, "ruby": 0.9, "rust": 0.95, "scala": 0.94, "scss": 0.95, "sevenzip": 0.95, "sgml": 0.95, "shell": 0.95, "smali": 0.95, "snap": 0.5, "solidity": 0.95, "sql": 0.9, "sqlite": 0.95, "squashfs": 0.95, "srt": 0.95, "stlbinary": 0.55, "stltext": 0.91, "sum": 0.95, "svg": 0.95, "swf": 0.95, "swift": 0.95, "tar": 0.95, "tcl": 0.95, "textproto": 0.95, "tga": 0.95, "thumbsdb": 0.95, "tiff": 0.95, "toml": 0.95, "torrent": 0.95, "tsv": 0.95, "ttf": 0.95, "twig": 0.95, "txt": 0.65, "typescript": 0.95, "unknown": 0.95, "vba": 0.85, "vcxproj": 0.95, "verilog": 0.95, "vhdl": 0.95, "vtt": 0.95, "vue": 0.95, "wasm": 0.95, "wav": 0.95, "webm": 0.95, "webp": 0.95, "winregistry": 0.95, "wmf": 0.95, "woff": 0.95, "woff2": 0.95, "xar": 0.95, "xls": 0.95, "xlsb": 0.95, "xlsx": 0.95, "xml": 0.92, "xpi": 0.95, "xz": 0.95, "yaml": 0.8, "yara": 0.5, "zig": 0.95, "zip": 0.91, "zlibstream": 0.95}, "overwrite_map": {}}
```
ia0, line 1: Note: This is not a copy of the existing file. Thresholds are being added.

invernizzi (approved):

assets/models/standard_v2/config.min.json
reyammer: Correct! we'll need to run the various syncs, including for python. will get to it. /cc @invernizzi

ia0:
We should have made the thresholds a different PR. They are wrong. A lot of files are not correctly identified. For example `tests_data/basic/json/doc.json` is low-confidence and considered text. Same for the documentation test of the Rust library which is (and not recognized as shell anymore due to threshold):

```shell
#!/bin/sh
echo hello
```

reyammer:
argh, sorry. you are right... I added the threhsolds in assets but didn't sync with python and didn't test, need to investigate what's going on.

PR #644, opened 2024-08-28
Update Rust tests to also run on previous_missdetections

ia0:
This is a follow-up of #632

### DIFF
--- a/rust/cli/output
+++ b/rust/cli/output
@@ -1,108 +1,108 @@
-+ magika test.sh
-test.sh: Shell script (code)
-+ magika test.sh --colors
-[1;34mtest.sh: Shell script (code)[0m
-+ magika test.sh --output-score
-test.sh: Shell script (code) 99%
-+ magika test.sh --json
++ magika rust/code.rs
+rust/code.rs: Rust source (code)
++ magika rust/code.rs --colors
+[1;34mrust/code.rs: Rust source (code)[0m
++ magika rust/code.rs --output-score
+rust/code.rs: Rust source (code) 89%
++ magika rust/code.rs --json
 [
   {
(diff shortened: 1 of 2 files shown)


rust/cli/output
```diff
@@ -100,9 +100,9 @@ test.sh: Shell script (code) 99%
   }
 ]
 + magika test.sh --jsonl
-{"path":"test.sh","result":{"status":"ok","value":{"dl":{"description":"Shell script","extensions":["sh"],"group":"code","is_text":true,"label":"shell","mime_type":"text/x-shellscript"},"output":{"description":"Shell script","extensions":["sh"],"group":"code","is_text":true,"label":"shell","mime_type":"text/x-shellscript"},"score":0.9980000257492065}}}
+{"path":"test.sh","result":{"status":"ok","value":{"dl":{"description":"Shell script","extensions":["sh"],"group":"code","is_text":true,"label":"shell","mime_type":"text/x-shellscript"},"output":{"description":"Shell script","extensions":["sh"],"group":"code","is_text":true,"label":"shell","mime_type":"text/x-shellscript"},"score":0.9990000128746033}}}
```
reyammer, line 103: did we ever figure out the source of this non-determinism?

reyammer (dismissed):

rust/cli/output
ia0: This is not non-determinism. I'm modifying the file (it's recursive, I use the test script itself as input). I could also use a file from the basic test suite, probably better.

ia0: Ok, I'm using `basic/rust/code.rs` and `basic/python/code.py` which hopefully shouldn't change often.

reyammer (approved):

PR #645, opened 2024-08-28
Use the new assets directory in Rust

ia0:
This is part of #640


ia0:
This PR is actually correct, the problem are the thresholds as  described in https://github.com/google/magika/pull/643#issuecomment-2315727306.

reyammer:
See PR #648 to temporary remove thresholds so we don't break existing tests.

reyammer (approved):

Issue #646, opened 2024-08-28
Removing `Cargo.lock` from `rust/lib`


tugunchinov:
Consider adding `Cargo.lock` to `.gitignore` and deleting it from sources. 
After that please update lib's dependencies:

```
[dependencies]
ndarray = "0.16.1"
serde = { version = "1.0.204", features = ["derive"], optional = true }
thiserror = "1.0.63"
tokio = { version = "1.38.1", features = ["fs", "io-util"] }

[dependencies.ort]
version = "2.0.0-rc.5"
default-features = true
features = ["ndarray"]
```

I'd like to use your crate but can't as it has fixed `Cargo.lock` but I have to build it from scratch and it doesn't compile.

ia0:
Did you try `cargo install --locked --path=rust/cli`? Note that we will soon make a release to crates.io, so you would be able to run `cargo install --locked magika-cli` instead.

tugunchinov:
Oh, I wanted to use `rust/lib` in my own project to check files' extensions.

ia0:
I see, in that case the lock file is not used. Soon you'll be able to use crates.io though.

tugunchinov:
Thank you, I'm looking forward to that

PR #647, opened 2024-08-28
Update Rust dependencies

tugunchinov:
For the #646


rust/.gitignore
```diff
@@ -1 +1,2 @@
 /target/
+/lib/Cargo.lock
```
ia0, line 2: Lock files are checked in for reproducibility purposes.

See https://github.com/rust-lang/cargo/blob/master/src/doc/src/faq.md#why-have-cargolock-in-version-control but this is indeed a contentious point in the Rust community and there are not yet clear guidelines. Note though that `cargo new` changed behavior to check the lock file, so this is the current default.

ia0 (changes requested):
Thanks! I'm happy to approve the dependency requirements update. But I wish to keep the lock file for reproducibility.

rust/.gitignore
tugunchinov: Okay, I think I had outdated information about keeping `Cargo.lock`. Thank you, I'll refresh my brain's database. 

Anyway, I pushed the commit with dependencies updates only.

ia0 (commented):
Thanks! You'll need to run `./test.sh` from the `rust` directory to update the CLI lock file too and make sure the CI passes.

tugunchinov:
Done 🫡

ia0 (approved):
Thanks! Running the CI.

PR #648, opened 2024-08-28
assets: temporary remove thresholds from standard_v2

reyammer:
thresholds need a second look and they are creating problems. Temporary removing them from assets to not break existing tests.


ia0 (approved):
Thanks!

invernizzi (approved):

PR #690, opened 2024-09-16
tests: add example for pytorch

reyammer:


reyammer:
This has to be merged after #691, as the old model doesn't support pytorch and the tests would fail.

PR #691, opened 2024-09-16
add standard_v2_1 model

reyammer:


reyammer:
This PR adds `standard_v2_1` model. The only differences wrt `standard_v2_0` are:
- new synthetic text dataset, which should help clearing some potential false positives
- support for pytorch
- updated thresholds for latex and pascal.

@ia0, a couple of points:
- not sure how you are handling rust's changelog, but I've noticed the rust's changelog checker script raising a warning without failing (I've noticed just by chance!). Not sure whether this "no failure" is intended. Regardless, remember to update the changelog for rust's -rc.1.
- the new model returns correct predictions for all test cases, but sometimes the score changes a bit. This seems to cause failures for rust's tests. I've already updated the model symlink and I've run `sync.sh`, but not sure how to update the "expected output".

ia0 (approved):

reyammer:
Thanks! and ack on the changelog thingy. Tricky. About `continue-on-error`, my understanding is that the job would be successful, but the UI would show a warning for the task that failed. But maybe that's annoying as well. One other thing we could consider is to have a changelog check script that is used when tagging a new version; the script could compare the changelog from the previous non-dev release and make sure there is at least one difference. But anyways, just food for thoughts, seems an overkill for now.

ia0:
> One other thing we could consider is to have a changelog check script that is used when tagging a new version

The problem with this is that once you need to update the changelog it's too late, you already forgot what you did and why you did it. The changelog should be updated with each and every change when everything is fresh. Doing it during release just makes releasing more difficult and provides worse changelogs. If we ever miss a warning, then we can try to fail the job and see how much the false positives are a pain.

PR #702, opened 2024-09-20
python: script to build python package

### LINKED ISSUE #566: Package the rust CLI within the python package

reyammer:
This implements a script to build the python package. While in theory this would just be `uv build`, we are hitting a number of problems / corner cases with maturin. This script implements a temporary solution to workaround all existing problems; we'll clean up as soon as the underlying tooling gets better.

Closes #566.

### DIFF
--- a/python/scripts/build_python_package.py
+++ b/python/scripts/build_python_package.py
@@ -1,0 +1,156 @@
+#!/usr/bin/env python
+# Copyright 2023-2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import re
+import subprocess
+import sys
+from pathlib import Path
+
+import click
+
+
+@click.command()
+def main() -> None:
+    repo_root_dir = Path(__file__).parent.parent.parent
+    python_root_dir = repo_root_dir / "python"
+    dist_output_dir = python_root_dir / "dist"
+    rust_root_dir = repo_root_dir / "rust"
+    # Compute paths to files we'll need to restore at the end of the build
+    rust_cli_readme_path = rust_root_dir / "cli" / "README.md"
+    rust_main_rs_path = rust_root_dir / "cli" / "src" / "main.rs"
+    rust_cli_cargo_toml_path = rust_root_dir / "cli" / "Cargo.toml"
+    rust_cli_cargo_lock_path = rust_root_dir / "cli" / "Cargo.lock"
+
+    r = subprocess.run(
+        ["git", "status", "--porcelain"], capture_output=True, cwd=repo_root_dir
+    )
+    if len(r.stdout) > 0 or len(r.stderr) > 0:
+        print("ERROR: the git repository is in a dirty state.")
+        sys.exit(1)
+
+    # delete readme as it causes issues with sdist; we'll restore it later
+    rust_cli_readme_path.unlink()
+
+    try:
+        # get the rust version from Cargo.toml and patch main.rs
+        rust_version = get_rust_version(rust_root_dir)
+        patch_main_rs_with_version(rust_main_rs_path, rust_version)
+
+        # get the python version from magika.__version__ and patch Cargo.toml
+        python_version = get_python_version(python_root_dir)
+        patch_cargo_toml_with_version(rust_cli_cargo_toml_path, python_version)
+
+        # update Cargo.lock
+        r = subprocess.run(
+            ["cargo", "check"],
+            cwd=rust_root_dir / "cli",
+        )
+
+        # build the package
+        r = subprocess.run(
+            ["uv", "build", "--out-dir", str(dist_output_dir)],
+            cwd=python_root_dir,
+        )
+        print(f"Python packages built at {dist_output_dir}")
+    finally:
+        # restore what we modified, no matter what happened
+        r = subprocess.run(
+            [
+                "git",
+                "restore",
+                "--",
+                str(rust_cli_readme_path),
+                str(rust_main_rs_path),
+                str(rust_cli_cargo_toml_path),
+                str(rust_cli_cargo_lock_path),
+            ],
+            capture_output=True,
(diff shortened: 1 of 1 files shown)


reyammer:
This is a quick POC to build the python script, it seems to work. The only problem I see is that the version number of the python package is taken from the rust package, and not from the python one. This is a problem as python packages have their own versions on pypi and we need to keep them separate.

ia0:
Interesting, this is https://github.com/PyO3/maturin/issues/2163 and I don't see any good workaround. What we could do is hard-code the `--version` of the CLI instead of it being automatically built by `clap` and modify the `Cargo.toml` version during build (like we do for the README). This is ugly but would unblock the situation. What do you think?

reyammer:
I guess that ugly is better that nothing for the time being? :-)

Instead of hardcoding the version, we could also patch the appropriate rust files as part of the build script?

To make things worse, the version of the python package is computed at run time (`pyproject.toml` is setup for "dynamic" version => the build system takes the version from well-known python variables, such as `magika.__version__` in our case).

Anyways, it would be nice to contain all the ugly parts in a specific place, and to keep as much as possible clean. So we could aim at keeping the codebase clean, and stashing all the bad stuff in a build script.

Now, if we go forward with something like this, is it clear to you how would the full build / publish package look like?

Things were much simpler in the past: I would build the package on my machine, and then publish it myself to pypi. Now, with the multiarch support for the rust binary, I guess the plan is to use gh actions to build the various versions of the package; gh actions would then produce some artifacts; and then? do we already know what's the well-lit path to publish the various wheels?

/cc @invernizzi

ia0:
The problem of using a hack is that it needs to work on Windows too, so it probably needs to be a Rust (or Python) program and not a shell script. But otherwise, I think it's possible to do the following:
- Replace the CLI `--version` with `Cargo.toml:package.version`
- Replace `Cargo.toml:package.version` with `__init.py__:__version__`
- Delete the CLI README
- Build
- Restore README.md and Cargo.toml

python/scripts/build_python_package.sh
```diff
@@ -0,0 +1,44 @@
...
+
+# From https://stackoverflow.com/a/246128
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+PYTHON_ROOT_DIR=$(realpath "$SCRIPT_DIR/..")
+DIST_OUTPUT_DIR=$PYTHON_ROOT_DIR/dist
+RUST_CLI_README_REL_PATH="rust/cli/README.md"
+RUST_CLI_README_ABS_PATH=$(realpath "$SCRIPT_DIR/../../$RUST_CLI_README_REL_PATH")
+
+# if rust's cli readme has local modifications, abort
+if [[ -n $(git status --porcelain | grep $RUST_CLI_README_REL_PATH) ]]; then
```
ia0, line 31: We could simply `[ -z "$(git status --porcelain)" ]` at the beginning of the script. We shouldn't run this if the repository is dirty right?

python/scripts/build_python_package.sh
```diff
@@ -0,0 +1,44 @@
...
+# rust/python setup currently has some issues and the building process fails if
+# rust's cli has a readme. Thus, we delete such readme (only if it is safe to do
+# so), we build the python package, and we then restore the readme.
+
+# From https://stackoverflow.com/a/246128
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+PYTHON_ROOT_DIR=$(realpath "$SCRIPT_DIR/..")
+DIST_OUTPUT_DIR=$PYTHON_ROOT_DIR/dist
+RUST_CLI_README_REL_PATH="rust/cli/README.md"
+RUST_CLI_README_ABS_PATH=$(realpath "$SCRIPT_DIR/../../$RUST_CLI_README_REL_PATH")
```
ia0, line 28: Why do you need absolute path?

reyammer:
> The problem of using a hack is that it needs to work on Windows too

Do we need this because we need to compile the rust binary for a windows env as well? I thought the complications were due to the different architectures (x86, arm, etc.), rather than OS?

Anyways, I'll rewrite it in python, it's not a super trivial script anymore.

python/scripts/build_python_package.sh
reyammer: true, but in practice it's annoying when developing this script itself as the git repo will always be in a dirty state. But it's a good point, will do this in my python rewrite.

ia0:
> Do we need this because we need to compile the rust binary for a windows env as well?

Yes, but that's not enough to explain why the script needs to work on Windows, because we could cross-compile. The reason we don't cross-compile is because of Maturin (I expect the `ort` crate to support cross-compilation correctly but didn't test either). Maybe there's a way to make it work but that's probably easier to have a portable script and keep the Maturin setup closer to what they assume.

python/scripts/build_python_package.sh
ia0: Yes, in those cases I just comment the check during development (but I also sometimes just create a fake commit with my development changes which is sometimes needed depending on what the script does).

reyammer:
Almost done, two questions in the meantime:
- in https://github.com/google/magika/pull/702#issuecomment-2363200374, the first step is `Replace the CLI --version with Cargo.toml:package.version`: can you confirm this will happen within the rust codebase?
- after I patch Cargo.toml, `uv build` complains with: `error: the lock file /magika-github/rust/cli/Cargo.lock needs to be updated but --locked was passed to prevent this`. Which command should I run to update the lock file? and, is it safe to do so?

ia0:
> * in [python: script to build python package #702 (comment)](https://github.com/google/magika/pull/702#issuecomment-2363200374), the first step is `Replace the CLI --version with Cargo.toml:package.version`: can you confirm this will happen within the rust codebase?

No, it's the script. You essentially need to change this line:

https://github.com/google/magika/blob/5f07dd1f6b3179bb31b75bbe538dcfddca548853/rust/cli/src/main.rs#L64

to something like this (where the version is taken from `Cargo.toml:package.version`):

```rust
        let binary = "0.1.0-rc.0-dev";
```

Feel free to change the initial line to prevent modifications:

```rust
        let binary = clap::crate_version!(); // keep in sync with python build script
```

> * after I patch Cargo.toml, `uv build` complains with: `error: the lock file /magika-github/rust/cli/Cargo.lock needs to be updated but --locked was passed to prevent this`. Which command should I run to update the lock file? and, is it safe to do so?

It should be enough to run `cargo check` from the `rust/cli` directory.

reyammer:
Success!

Now the resulting python package has the correct version, and the magika rust binary has its own (correct) version as well.

I'll wait for your quick review before merging, thanks!

ia0 (approved):

PR #704, opened 2024-09-20
Fix CLI name in --version

ia0:
The default is the package name which is `magika-cli`. We want to use the binary name which is `magika`. Feature request on `clap` side: https://github.com/clap-rs/clap/issues/5738.

### DIFF
--- a/rust/cli/CHANGELOG.md
+++ b/rust/cli/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ### Patch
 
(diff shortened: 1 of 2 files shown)


reyammer (approved):

reyammer:
Just to check, the feature request for clap folks is just to change the default, and in this PR you implemented the desired behavior yourself?

ia0:
> Just to check, the feature request for clap folks is just to change the default, and in this PR you implemented the desired behavior yourself?

Yes correct.

PR #715, opened 2024-09-23
python: improve API usability and add backward compatibility layer

reyammer:
This PR makes progress on two aspects:
- simplify access to Magika's API results
- implement a backward compatibility layer, add deprecation warnings, and error out with meaningful messages when we cannot be fully backward compatible.

More context in the description of the two commits.

### DIFF
--- a/python/src/magika/__init__.py
+++ b/python/src/magika/__init__.py
@@ -19,10 +19,10 @@ __version__ = "0.6.0-dev0"
 import dotenv
 
 from magika import magika
-from magika.types import prediction_mode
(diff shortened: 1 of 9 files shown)


invernizzi (dismissed):

invernizzi:
LGTM. I'd consider using [`DeprecationWarning`](https://docs.python.org/3/library/exceptions.html#DeprecationWarning), like [so](https://dev.to/hckjck/python-deprecation-2mof), so that people can filter the warnings if they choose to.

PR #717, opened 2024-09-25
gh actions: improve python testing

reyammer:
Improve the python gh action tests:
- clean up the gh action that runs the test suite.
- improve e2e gh action: it builds the package, it installs the package with pip, it runs the magika CLI (the rust one) and the Magika python module against all basic samples and checks for correct predictions.

### DIFF
--- a/.github/workflows/python-e2e-test.yml
+++ b/.github/workflows/python-e2e-test.yml
@@ -20,8 +20,10 @@ jobs:
   e2e-testing:
     strategy:
       matrix:
-        python-version: [ "3.8.x", "3.9.x", "3.10.x", "3.11.x", "3.12.x" ]
-        os: [ "ubuntu-latest", "macos-latest", "windows-latest" ]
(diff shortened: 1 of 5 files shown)


.github/workflows/python-e2e-test.yml
```diff
@@ -29,23 +33,37 @@ jobs:
...
         shell: bash # Allows for cross-platform
 
+      - name: Run magika --version
+        working-directory: python
+        run: magika --version
+        shell: bash # Allows for cross-platform
+
       - name: Run magika with tests_data
-        run: magika -r tests_data/ || exit 1
+        run: magika -r tests_data/basic || exit 1
```
ia0, line 63: nit: `|| exit 1` should be redundant, GitHub should run with `set -e`.

https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell

ia0 (approved):

.github/workflows/python-e2e-test.yml
reyammer: fixed

ia0 (approved):

PR #718, opened 2024-09-25
Release Rust crates

ia0:


reyammer (approved):

ia0:
Need to fix MacOS.

ia0:
MacOS "fix" coming in #719. Not sure what's the default in ONNX, but seems like we need to set some value.

reyammer:
Note that the checks didn't pick up yet the new gh actions workflow that actually builds & test the package. Can we re-trigger all the workflows?

ia0:
> Note that the checks didn't pick up yet the new gh actions workflow that actually builds & test the package. Can we re-trigger all the workflows?

The new workflow doesn't trigger when only rust files are touched, so it's working as intended. Why do you want the workflow to run? Maybe removing the paths restriction would be better?

reyammer:
good catch; I thought it was just stale. I just updated the e2e test workflow to trigger also for rust/** modifications. Now it's running the checks again, and the build package is triggered, seems to work.

ia0:
> good catch; I thought it was just stale. I just updated the e2e test workflow to trigger also for rust/** modifications. Now it's running the checks again, and the build package is triggered, seems to work.

Sounds good. I'll fix the problem in a different PR (to keep this PR small). It's a bug in the recent nightly https://github.com/rust-lang/rust-clippy/issues/13458

ia0:
This is tagged as [rust-v0.1.0-rc.1](https://github.com/google/magika/releases/tag/rust-v0.1.0-rc.1) and published as [magika](https://crates.io/crates/magika/0.1.0-rc.1) and [magika-cli](https://crates.io/crates/magika-cli/0.1.0-rc.1). Because it's a pre-release, the version must be explicitly specified:

| Crate | Command to use or install |
| --- | --- |
| Library | `cargo add magika@=0.1.0-rc.1` |
| CLI | `cargo install --locked magika-cli@0.1.0-rc.1` |

Issue #724, opened 2024-09-25
Add gh workflow to build python packages for all platforms and publish as artifacts


reyammer:
Steps:
- [x] Disable maturin workflow via github UI; we build via `./python/script/build_python_package.py`
- [x] Disable running python-related workflows when we tag a new version of rust
- [x] Create a new workflow that builds & publish python package artifacts to github when we tag a python version
- [x] Once we are done porting the maturin workflow to our setup, delete the maturin workflow

@ia0, am I missing anything?

ia0:
> * [ ]  Disable running python-related workflows when we tag a new version of rust

Should theoretically not be needed since the tag name won't match those workflows AFAICT. We tag rust as `rust-v*`.

LGTM otherwise.

PR #728, opened 2024-09-26
gh action: build python packages for different archs

reyammer:
Closes: #724.

Build wheels/tarball for linux, windows, and mac. Then, upload the files as artifacts.


.github/workflows/python-build-and-upload-packages.yml
```diff
@@ -0,0 +1,63 @@
...
+        working-directory: python
+        run: uv sync --all-extras --dev
+
+      - name: Build the python package
+        working-directory: python
+        run: uv run ./scripts/build_python_package.py
+
+      - name: List wheels
+        working-directory: python
+        run: python -c "import os; print('\n'.join(os.listdir('dist')))"
```
ia0, line 57: It seems like we create the source wheel for all runners, thus using 3 times more space as needed. Is there a way to use `uv build` to only build either the source or binary package? We could then have 4 configs in the matrix: linux, macos, windows, source. And build 4 artifacts that we can unzip in the same directory without overwriting a tar-ball with a similar tar-ball.

.github/workflows/python-build-and-upload-packages.yml
```diff
@@ -0,0 +1,63 @@
...
+
+on:
+  push:
+    branches:
+      - 'main'
+    tags:
+      - 'python-v*'
+  pull_request:
+    # In theory we wouldn't need to create artifacts for each PR, but this makes
+    # it much easier to test this workflow and there are no big cons.
```
ia0, line 11: If the goal was only testing within this PR, then it's also possible to just comment `pull_request` before merging.

ia0 (dismissed):

.github/workflows/python-build-and-upload-packages.yml
reyammer: true. instead of making the build python package script more complicated, for now I modified the workflow to upload the tarball only for linux. This has also the advantage of having 3 runners instead of 4. it's not ideal on some aspects, I propose to fix later if needed.

.github/workflows/python-build-and-upload-packages.yml
reyammer: agreed. updated the comment to reflect this. and will comment that part before merging.

.github/workflows/python-build-and-upload-packages.yml
ia0: That's also a good solution. I don't think we need to improve it.

ia0 (approved):
Will reapprove again once the `pull_request` trigger is commented.

reyammer:
FYI: the maturin workflow was using the wrong syntax for `if`; with that syntax, the`if` always evaluates to `true`, leading to problems. Diff: https://github.com/google/magika/compare/6a9ac7bdb94363e66b2cbb0f10df973bfd044762..cc90211e0397a2d6699ca6fdedcc570abeb75a43

Should be good to go now.

reyammer:
Just realized I need to comment out the full pull_request block, not just the content...

ia0 (approved):

PR #739, opened 2024-10-02
Add function to check python readme links are valid and absolute

### LINKED ISSUE #730: Add GH workflow to check that the python package README does not contain relative paths
`./python/README.md` is what is rendered at https://pypi.org/project/magika/. Thus, we need to make sure that links to files within this repo are absolute link (with the full https://github.com/...) rather than relative links. Also check that such links are valid links.

brokoli777:
I added a check for the python README file to check if all the links are valid and not relative. 

If the links are not valid, it will print out the invalid links for review and stops the script (stopping the package build in the process).

Currently, this is assuming that all the links in the README file will be pointing to GitHub, if that is not the case, I can modify it to check if it is an HTTPS link.

Closes #730.


python/scripts/build_python_package.py
```diff
@@ -33,6 +34,10 @@ def main() -> None:
     rust_cli_cargo_toml_path = rust_root_dir / "cli" / "Cargo.toml"
     rust_cli_cargo_lock_path = rust_root_dir / "cli" / "Cargo.lock"
 
+    # Check python README links
+    python_readme_path = python_root_dir / "README.md"
+    check_readme_links(python_readme_path)
```
reyammer, line 39: let's have a more explicit name for this function, e.g., "check_readme_has_absolute_links".

python/scripts/build_python_package.py
```diff
@@ -151,5 +156,35 @@ def patch_line_matching_prefix(file_path: Path, prefix: str, new_line: str) -> N
...
+    """Check if the README contains valid absolute links."""
+    readme_content = readme_path.read_text()
+    
+    # Find all markdown links
+    link_regex = r'\[.*?\]\((.*?)\)'
+    links = re.findall(link_regex, readme_content)
+
+    invalid_links = []
+    for link in links:
+        if not link.startswith("https://github.com/"):
```
reyammer, line 169: let's be more generic, just check they start with https://. We may have links to external blogposts, etc.

python/scripts/build_python_package.py
```diff
@@ -151,5 +156,35 @@ def patch_line_matching_prefix(file_path: Path, prefix: str, new_line: str) -> N
...
+            invalid_links.append(f"Relative link found: {link}")
+        else:
```
reyammer, line 171: it feels more readable if the logic is inverted: "if the link starts with https" => check it's valid; "else: add link to invalid link"

python/scripts/build_python_package.py
```diff
@@ -151,5 +156,35 @@ def patch_line_matching_prefix(file_path: Path, prefix: str, new_line: str) -> N
...
+            # Check if the link is valid
+            try:
+                response = requests.head(link, allow_redirects=True, timeout=5)
+                if response.status_code != 200:
+                    invalid_links.append(f"Invalid link: {link} (status code: {response.status_code})")
+            except requests.RequestException as e:
+                invalid_links.append(f"Error accessing link: {link} ({e})")
+    
+    if invalid_links:
```
reyammer, line 180: imo is more readable to have `if len(invalid_links) > 0`

python/scripts/build_python_package.py
```diff
@@ -151,5 +156,35 @@ def patch_line_matching_prefix(file_path: Path, prefix: str, new_line: str) -> N
...
+        print("README.md contains invalid or relative links:")
+        for error in invalid_links:
+            print(error)
+        sys.exit(1)
+    else:
```
reyammer, line 185: no need for this else, just print that all is good.

python/scripts/build_python_package.py
```diff
@@ -151,5 +156,35 @@ def patch_line_matching_prefix(file_path: Path, prefix: str, new_line: str) -> N
...
+        print("All links in README.md are valid.")
```
reyammer, line 186: Replace "README.md" with `readme_path`

reyammer (commented):
Please do pass with `ruff` on this script (see https://github.com/google/magika/actions/runs/11150217299/job/31080639258?pr=739)

I suggest to setup ruff to be run "on save" in your IDE (eg vscode).

python/scripts/build_python_package.py
```diff
@@ -151,5 +156,35 @@ def patch_line_matching_prefix(file_path: Path, prefix: str, new_line: str) -> N
     file_path.write_text("\n".join(lines))
 
 
+def check_readme_links(readme_path: Path) -> None:
+    """Check if the README contains valid absolute links."""
```
reyammer, line 160: Improve docs: mention that the function makes the script exit with error code in case of problems.

python/scripts/build_python_package.py
reyammer: actually, I propose to make this function more generic: `check_markdown_has_only_absolute_links`. then let's remove the references about "Readme" in the body of this function, as it's now about markdown in general.

reyammer:
Thanks for the PR! Did a first review. Let me know if you have questions.

reyammer:
Hello @brokoli777, still interested in helping out on this or should I take from here?

brokoli777:
Oh very sorry! For some reason, I didn't realize there was a reply to the pull request. I will do the recommended changes right now.

python/scripts/build_python_package.py
brokoli777: Good point, I updated the code.

python/scripts/build_python_package.py
brokoli777: I agree, I update the code

python/scripts/build_python_package.py
brokoli777: Got it! I also updated the function to check for either directories or a file based on the file path to be more generic.

reyammer:
Thanks, new round of comments:
- ruff pass still fails, can you double check?
- in case of errors, it seems the script just prints errors and does not really exit with 1?
- I saw that the new function takes as input a `markdown_path` (which indicates it's a path to a given file), but the code tries to check whether it's a directory and does some recursive check to find new markdown. I find this confusing, I propose to keep it simple: the function should take a `markdown_path` (as it is now), and this should just be a path to a given markdown file; let's remove directory checks or anything like this. Unneeded complexity. So, let's move the code from the inner function to the outer function and then remove the inner function itself. and let's make sure it does exit with 1 in case of problems. Let's also update the comment to reflect that (focus on one markdown, not one or more).

Thanks!

reyammer (approved):

reyammer:
Merged, thanks!

PR #746, opened 2024-10-10
python: add twine as dev dep

reyammer:


python/uv.lock
```diff
@@ -215,9 +476,37 @@ wheels = [
...
+]
+sdist = { url = "https://files.pythonhosted.org/packages/a5/1c/2bdbcfd5d59dc6274ffb175bc29aa07ecbfab196830e0cfbde7bd861a2ea/keyring-25.4.1.tar.gz", hash = "sha256:b07ebc55f3e8ed86ac81dd31ef14e81ace9dd9c3d4b5d77a6e9a2016d0d71a1b", size = 62491 }
+wheels = [
+    { url = "https://files.pythonhosted.org/packages/83/25/e6d59e5f0a0508d0dca8bb98c7f7fd3772fc943ac3f53d5ab18a218d32c0/keyring-25.4.1-py3-none-any.whl", hash = "sha256:5426f817cf7f6f007ba5ec722b1bcad95a75b27d780343772ad76b17cb47b0bf", size = 38946 },
+]
+
 [[package]]
 name = "magika"
-version = "0.6.0.dev0"
+version = "0.1.0rc1"
```
reyammer, line 509: FYI: @ia0, the maturin setup is messing up the uv.lock magika's python version as well... should not be an actual problem as this should just mean that, locally, the system thinks it has magika 0.1.0rc1 installed, but that doesn't really matter....

ia0: Yes this sounds exactly like https://github.com/PyO3/maturin/issues/2163

PR #749, opened 2024-10-10
Update python-build-and-upload-packages.yml

reyammer:
Build linux package on old ubuntu


reyammer:
FYI @ia0, hopefully this make sure that the package is compatible with ubuntu 20.04 forward.

ia0:
Sounds good, thanks!

PR #779, opened 2024-11-01
Improve python github workflows

reyammer:
This PR consolidates the python-related workflows and package build scripts:
- `.github/workflows/python-test-suite.yml` now tests everything it can that does NOT need the wheels. This includes testing the magika rust binary and the magika python package.
- `maturin.yml` is now THE GH workflow in which we build the wheels, test them, and upload them as artifacts.
- removed all other python-related GH workflows, as the python packages/wheels are now created and tested in `maturin.yml`.
- removed the `build_python_package.py` script; we build the wheels via `maturin.yml`.


reyammer:
The resulting linux wheel is still not pip installable on ubuntu 20.04 (see https://github.com/google/magika/issues/781), but this PR is a good clean up anyways.

PR #792, opened 2024-11-05
Added Coverage Analysis for Tests of Python Package #221 and Update Tox and Pytest Configurations

ps2181:
This PR updates the tox and pytest configuration files to improve test coverage reporting and streamline the testing process. The changes include:
tox.ini:
pytest.ini

Added coverage commands to track test coverage.
Defined separate test environments for running tests and coverage.
Configured pytest to report coverage in HTML format.

A terminal output showing the success or failure of tests.
An HTML coverage report accessible via htmlcov/index.html, detailing the coverage statistics for the project.
you can run the coverage report my either using the tox command or the pytest command where it generates a html report in htmlcov under index.html
#221


reyammer:
Hello, sorry for the delay in getting back to you.

Some feedback:
- the PR seems to contain a lot of very unrelated modifications (tox, gitignore, an executable called `da`?) Can you clean that up? Note: it's important that the PR only has one commit (so, you should squash; or create a new PR if you are unsure on how to do that) that touches the right files; otherwise these unrelated modifications would be part of the git history
- ideally we have something like https://github.com/google/scaaml/blob/main/.github/workflows/pytest.yml#L72 (/cc @kralka), integrate with coverall, so that we get a coverage report (see https://github.com/google/scaaml/pull/325 as example)

reyammer:
As per #837, closing this one.

PR #825, opened 2024-12-09
Extend magika wheels test to windows runner

### LINKED ISSUE #805: Extend test "uv add magika.whl" to windows runner
We recently updated our python github workflows to test that the magika package/wheels can be installed with `uv add` (in addition to `pip install magika`). To test this, we use a one-liner that currently works only for ubuntu and macos. We need to extend the one-liner (or add another one) so that it also works for Windows.

These are the workflows to update:
- [ ] https://github.com/google/magika/blob/main/.github/workflows/python-test-published-package.yml
- [ ] https://github.com/google/magika/blob/main/.github/workflows/python-test-published-rc-package.yml
- [ ] https://github.com/google/magika/blob/main/.github/workflows/python-build-package.yml

Mohit-Gaur:
Resolves #805 

Updated the following workflows to install `uv` and verify that `uv add magika.whl` works on windows runner:

* `.github/workflows/python-build-package.yml`
* `.github/workflows/python-test-published-package.yml`
* `.github/workflows/python-test-published-rc-package.yml`

### DIFF
--- a/.github/workflows/python-build-package.yml
+++ b/.github/workflows/python-build-package.yml
@@ -61,10 +61,7 @@ jobs:
         name: Check that `uv add magika.whl` works
         run: mkdir /tmp/test-uv && cp -vR dist/*.whl /tmp/test-uv && cd /tmp/test-uv && uv init && uv add ./$(\ls -1 *.whl | head -n 1)
       - if: matrix.platform.runner == 'windows-latest'
-        name: Install uv on Windows
-        run: powershell -Command "Invoke-WebRequest -Uri https://astral.sh/uv/0.4.7/install.ps1 -OutFile install.ps1; ./install.ps1"
-      - if: matrix.platform.runner == 'windows-latest'
-        name: Check that `uv add magika.whl` works on Windows
+        name: Check that magika install with uv works on Windows
         shell: pwsh
         run: |
           mkdir C:\test-uv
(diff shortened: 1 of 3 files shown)


.github/workflows/python-build-package.yml
```diff
@@ -60,6 +60,19 @@ jobs:
       - if: matrix.platform.runner != 'windows-latest'
         name: Check that `uv add magika.whl` works
         run: mkdir /tmp/test-uv && cp -vR dist/*.whl /tmp/test-uv && cd /tmp/test-uv && uv init && uv add ./$(\ls -1 *.whl | head -n 1)
+      - if: matrix.platform.runner == 'windows-latest'
+        name: Install uv on Windows
```
reyammer, line 64: do we really need this? Line 37 should already install uv on windows as well? (same for the other two files)

Mohit-Gaur: As I understand it, line 37 uses a shell script which will not work on windows. That's why I added a separate step that uses a powershell script to handle uv install.

reyammer: From https://github.com/google/magika/actions/runs/12232019796/job/34140885923, it seems that it does run successfully (the command is run for sure as there is no `if` on platform runner OS; question is whether your powershell part can use the just-installed uv). The problematic part was the one-liner starting with `mkdir`, which your PR seems to address. Can you maybe try to remove the "install uv on windows" part and check that it still works?

Mohit-Gaur: Done! I removed that step and it still works fine

reyammer (approved):

Issue #662, opened 2024-08-30
Add more "basic" tests samples to cover supported content types


reyammer:
The new models ("standard_v2_x" and "standard_v3_0") supports 200+ content types: https://github.com/google/magika/tree/main/assets/models/standard_v3_0/README.md

Ideally, we have at least one "basic sample" for each of the supported content types (See `/tests_data/basic/*`).

This issue acts as a call for action -- external help is very welcome!

Important aspects to keep in mind:
- Content types for which we have no samples yet should be prioritized. Among these, prioritize more common content types rather than niche ones.
- The "basic" test samples (in the `tests_data/basic/<content_type>/*`) are supposed to be "easy to recognize". In other words, the goal for these samples is to check that the model does a reasonable job with clear-cut samples, rather than corner-cases.
- It's OK to group a bunch of test cases in a single PR.
- The PR should state the origin of each sample.
- The samples should NOT be taken from existing projects / online resources (in these settings, it would be very challenging to properly document the origin of these files); they should be manually written/created by the PR author.

mamamia96:
I'd like to add a handful of basic tests for:
- pickle
- powershell
- ttf
- gif

reyammer:
These would be very welcome! As indicated in the issue, please include a description on how these files were created (especially for the binary ones, such as pickle). Examples on how we created some of the test cases: create a new google doc, then "export as" various formats. Thanks!

mamamia96:
Where should I include my description of how I created the files?

dukecat0:
<details>
<summary>I made a list of samples added for further reference:</summary>

| #   | Content Type Label       | Added |
|-----|--------------------------|--------|
| 1   | 3gp                      |        |
| 2   | ace                      |        |
| 3   | ai                       |        |
| 4   | aidl                     |        |
| 5   | apk                      |        |
| 6   | applebplist              |        |
| 7   | appleplist               |        |
| 8   | asm                      | ✓      |
| 9   | asp                      |        |
| 10  | autohotkey               |        |
| 11  | autoit                   |        |
| 12  | awk                      |        |
| 13  | batch                    | ✓      |
| 14  | bazel                    |        |
| 15  | bib                      |        |
| 16  | bmp                      |        |
| 17  | bzip                     |        |
| 18  | c                        | ✓      |
| 19  | cab                      |        |
| 20  | cat                      |        |
| 21  | chm                      |        |
| 22  | clojure                  |        |
| 23  | cmake                    |        |
[...]

PR #840, opened 2024-12-28
feat: add rust to basic test_data with asm! macro

0xb-s:


reyammer:
Thanks for this and the other PRs about the examples! If you still have time, I would try to prioritize basic examples for content types we don't have examples yet (rather than variants of content types for which we do already have basic examples, such as this PR). See https://github.com/google/magika/issues/662 for more guidelines. Thanks!

PR #894, opened 2025-01-21
Disable some warnings when building ONNX Runtime

ia0:
GCC 14 is either better at detecting uninitialized issues or has false positives, compared to GCC 13. Those warnings are considered as errors, so we disable them.


ia0:
The problem is the manylinux container: https://github.com/pypa/manylinux/pull/1730

ia0:
It might be simpler to figure out how to remove `-Werror=maybe-uninitialized`.

reyammer (approved):

reyammer:
Thanks for tracking this down!

PR #900, opened 2025-01-21
Disable automatic git CRLF conversion at checkout

### LINKED ISSUE #899: Fix github workflows so that git checkout does not mess up CRLF on windows
TIL: git checkout patches newlines in different ways depending on the OS. On windows, "\n" are converted to "\r\n".

Fix the github workflows so that we disable this behavior (which messes up tests).

reyammer:
Fixes #899.

FYI @ia0 @invernizzi

### DIFF
--- a/.github/workflows/js-publish.yml
+++ b/.github/workflows/js-publish.yml
@@ -11,6 +11,8 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
+      - name: Disable automatic git CRLF conversion
+        run: git config --global core.autocrlf false
       - uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # pin@v4
       - uses: actions/setup-node@1e60f620b9541d16bece96c5465dc8ee9832be0b # pin@v4
         with:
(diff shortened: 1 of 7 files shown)


ia0:
It's probably better to use `.gitattributes` (at the repository root) for that, otherwise developers on Windows still have the issue.

```
# Make sure test files are checked-in verbatim.
/tests_data/** -text
```

ia0:
But this issue is still interesting, because it raises those questions:
- Is it a valid Python source to use CRLF instead of LF? If yes, do we have such training data (to avoid misclassification on Windows)?
- What about other text formats? The same questions apply.

reyammer:
To your questions:

> Is it a valid Python source to use CRLF instead of LF? If yes, do we have such training data (to avoid misclassification on Windows)?

we don't have a specific dataset that makes sure we have CRLF, but given the size of the dataset, I would expect some samples to have CRLF. And the fact that this bug has been there all along without being detected makes me think we should be OK. This feels just an unlucky FP with the new model (it never affected any other file for any of the 5 models we tested so far). But definitively something to keep in mind. If we get some feedback that this is causing actual issues, we should be able to add some augmentation and train for that as well.

> What about other text formats? The same questions apply.

Same as above; we didn't do anything specific about it. But so far it doesn't seem a big deal.

ia0:
Sounds good. And reading through the lines I guess it is valid to use CRLF in Python.

Issue #870, opened 2025-01-19
Discussion on having a source distribution that is pure python


reyammer:
One concern about the new 0.6.x version is that the python package is not pure python anymore (as the CLI is now in rust).

I think it makes sense to support as many platforms we can and upload wheels with the python codebase + rust client (as we are doing already).

But would it be possible to create a source-only distribution that just has the python codebase and no rust client? In this way, most clients would still have the benefits from the rust client (when they find a wheel that matches the requirements), and the clients that don't have a good match would not have the rust client, but they would still be able to install the python-only codebase (we could either just ship the python module without any client, or we can include a fallback basic implementation of a python client as a placeholder)?

Thoughts?

/cc @ia0

reyammer:
It seems we have two paths:

- generate a wheel that is pure python and with less specific tags so that pip and other resolvers consider it with low priority.
- generate a source distribution, which would contain the source and instructions on how to build the package (which pip would be in charge of)

In both cases, we would need to tweak the build workflow to remove the rust part and add the fallback python client.

reyammer:
Discussed with @ia0, we'll go for the generic wheel and see what happens.

reyammer:
Implemented and released 0.6.1rc0, which also has a pure python wheel: https://pypi.org/project/magika/0.6.1rc0/. We'll add a source distribution as well later on (See #919).

PR #909, opened 2025-01-22
Build pure python wheel

reyammer:
Context: #870.


.github/workflows/python-build-package.yml
```diff
@@ -76,35 +76,27 @@ jobs:
...
-  sdist:
+  build-pure-python-wheel:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - run: python3 ./python/scripts/fix_package_version.py
-      - name: Build sdist
-        uses: PyO3/maturin-action@v1
+      - run: uv run ./scripts/prepare_pyproject_for_pure_python_wheel.py
+          working-directory: python
```
ia0, line 92: I would expect less indentation here.
```suggestion
        working-directory: python
```

.github/workflows/python-build-package.yml
```diff
@@ -76,35 +76,27 @@ jobs:
...
+      - name: Build pure python wheel
+        run: uv build --wheel
         with:
```
ia0, line 95: I don't think this should be in a `with`. You can have `working-directory` directly at the same level as `run` (i.e. in a step).

reyammer: Indeed; I copied it from some maturin action, will fix. same for above.

.github/workflows/python-build-package.yml
reyammer: 👍

ia0 (dismissed):

.github/workflows/python-build-package.yml
```diff
@@ -76,35 +76,48 @@ jobs:
...
+        run: curl -LsSf https://astral.sh/uv/0.5.22/install.sh | sh
+      - run: uv run ./scripts/prepare_pyproject_for_pure_python_wheel.py
+        working-directory: python
+      - name: Build pure python wheel
+        run: uv build --wheel --out-dir ../dist
+        working-directory: python
+      - if: matrix.platform.runner != 'windows-latest'
+        name: Check that `uv add magika.whl` works
+        run: mkdir /tmp/test-uv && cp -vR dist/*.whl /tmp/test-uv && cd /tmp/test-uv && uv init && uv add ./$(\ls -1 *.whl | head -n 1)
+      - if: matrix.platform.runner == 'windows-latest'
```
ia0, line 101: There's no matrix here. I'm not sure why this is needed.

.github/workflows/python-build-package.yml
```diff
@@ -76,35 +76,48 @@ jobs:
...
-      - run: python3 ./python/scripts/fix_package_version.py
-      - name: Build sdist
-        uses: PyO3/maturin-action@v1
-        with:
-          command: sdist
-          args: --out=../dist
-          working-directory: python
-      - name: Upload sdist
+      - name: Install uv
+        run: curl -LsSf https://astral.sh/uv/0.5.22/install.sh | sh
```
ia0, line 92: The uv version can also be updated above.

.github/workflows/python-build-package.yml
reyammer: removed

.github/workflows/python-build-package.yml
reyammer: done

reyammer:
A bigger problem for the new pure python wheel is that there are close to no tests. I mean, it's only tested on that ubuntu runner, but ideally we would check that it somehow would work in the windows and macos runners. The trick seems to be: upload artifact, then in another job in the same workflow you download the artifact (in various runners) and run the tests there. I'll do that in another PR.

Issue #930, opened 2025-01-27
Make python client available in binary wheels, and change its name to `magika-python-client`


reyammer:
This is the current situation:
- binary wheel: `magika` => rust client. There is not python client.
- pure python wheel: `magika` => python client. There is no rust client.

This is not ideal:
- if for any reason pip selects the pure python instead of the rust one, clients will inadvertely use the python one, which has somewhat the same features, but is slower and its output is not exactly the same
- it is handy to have the python client in the binary wheel as well, as I often use it for quick tests (and, for the python client is trivial to support v1 and other versions; I could keep support for old versions around and use to compute some metrics. We would of course only ship in the package the most recent model, but the python client can easily be pointed to any model in the assets folder).

I propose to do the following:
- binary wheel: `magika` => rust client. `magika-python-client` => python client.
- pure python wheel: `magika` => print a warning that this is a pure python wheel with no rust, and that one should either install the proper wheel, or one can fall back to using the python client, with `magika-python-client`. `magika-python-client` => python client

@ia0, thoughts?

reyammer:
Closing this one as #931 is a pretty good fix already.

PR #931, opened 2025-01-29
Make the python client available as "magika-python-client" rather than "magika"

### LINKED ISSUE #919: Generate proper source distribution for python package
We were generating an sdist package via maturin, but it was somewhat broken, as it was including the rust binary (which should not be there for a source-only release). Adapt `build-pure-python-wheel` job from https://github.com/google/magika/blob/main/.github/workflows/python-build-package.yml and generate a new sdist artifact (likely as part of the same job).

reyammer:
Also, in the pure python wheel, "magika" now points to a small script that raises a warning and provide instructions.
Ref: #930.

Also: generate sdist alongside the pure python wheel. Fixes #919.

### DIFF
--- a/.github/workflows/python-build-package.yml
+++ b/.github/workflows/python-build-package.yml
@@ -224,8 +224,13 @@ jobs:
           uv add ".\$wheel"
       - name: Install the wheel
         run: python3 -m pip install $(python -c "import glob; print(glob.glob('dist/*.whl')[0])")
-      - run: magika --version
+
+      # Check that the magika script points to the placeholder raising a warning
+      - run: magika --version | grep -C10 WARNING | grep -C10 magika-python-client
+      # Check that the fallback magika's python client can be run
+      - run: magika-python-client -r tests_data/basic
+      # Check that the results of the python's client are correct
+      - run: python3 ./python/scripts/run_quick_test_magika_cli.py --client-path magika-python-client
+      # Test the python module
       - run: "python3 -c 'import magika; m = magika.Magika(); print(m)'"
-      - run: magika -r tests_data/basic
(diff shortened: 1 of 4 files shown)


.github/workflows/python-build-package.yml
```diff
@@ -57,32 +57,6 @@ jobs:
           target: ${{ matrix.platform.target }}
           args: --release --out=../dist
           working-directory: python
-      - if: matrix.platform.runner != 'windows-latest'
```
reyammer, line 60: Removed these tests as anyways we test all the wheels with various settings in the job below. Same for the pure wheels.

ia0: I fail to see which job. I see one called `test-pure-python-wheel` but it seems to only be the pure python wheel? In particular in only depends on `build-pure-python-wheel-and-sdist` and not on `build-wheels`.

ia0 (approved):
LGTM modulo the fact that I'm not sure I follow the fact that all wheels are tested.

.github/workflows/python-build-package.yml
reyammer: It's totally possible I screwed something up, so let's double check.

There are 4 jobs in total.

`test-wheels` is the job that test the wheels produced by `build-wheels`. All things I've removed from `build-wheels` were tests that we do anyways in `test-wheels` (with more platforms).

Then, we have `build-pure-python-wheel-and-sdist`: this builds and pure python wheel (and now also sdist; I just added it). The wheel is tested in the `test-pure-python-wheel` job. The sdist is currently untested (tracked: https://github.com/google/magika/issues/933). All tests I've removed from `build-pure...` are done anyways in `test-pure...`.

PR #979, opened 2025-03-11
Python: add new `identify_stream` API

### LINKED ISSUE #970: Implement: identify_stream(self, byte_stream: BytesIO) ...
Magika currently supports`identify_path` for file paths, and `identify_bytes` for byte strings. However, there are many cases where I already have an open file-like binary stream (e.g. from a prior open(...)). In this case, if the stream is seekable, it should be possible to support it efficiently, without reading it all into memory, similar identify_path.

Proposed approach would be:

```python
def identify_stream(self, byte_stream: BytesIO) -> MagikaResult:
       if not byte_stream.seekable():
           raise TypeError(
               f"Input byte_stream should be seekable."
           )
       ... Do your thing
```

reyammer:
This PR adds a new `identify_stream` API, as suggested in #970. With this API, one can identify the content type of an already-open stream; the implementation does not read the full file, and it just `seek()`s around. This new API avoids clients to be forced to (directly or indirectly) read the full file's content in memory.

Fixes #970.

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,7 +223,7 @@ Options:
 
 While the command line client is handy for many use cases, it may not be the most suitable for automated workflows. Thus, Magika comes with bindings for Python and other languages (see the [Bindings section](#bindings) below for more details).
 
-Here is a couple of examples on how to use the `Magika` Python module:
+Here is a few examples on how to use the `Magika` Python module:
 
 ```python
 >>> from magika import Magika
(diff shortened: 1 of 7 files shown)


afourney (commented):
I've reviewed the implementation, and I love the care to detail on common errors etc. I've not had a chance to test this yet, but will ASAP. I look forward to this feature.

PR #991, opened 2025-03-17
Release Rust crates

ia0:


reyammer:
LGTM, thanks!

Question: I guess you change the .toml, but then you run some script / update to sync the .lock files, right? how do you do that? (for now I just blindly run the `sync.sh` script, but is there a better way?)

ia0:
> Question: I guess you change the .toml, but then you run some script / update to sync the .lock files, right? how do you do that? (for now I just blindly run the `sync.sh` script, but is there a better way?)

No, I just blindly remove the `-dev` suffix from the `Cargo.{toml,lock}` and `CHANGELOG.md`, see the `rust/publish.sh` script. I don't know how to update the .lock files without at least `cargo metadata`.

ia0:
It's published on crates.io.

PR #998, opened 2025-03-17
docs: add rust docs tracking to check docs script

### LINKED ISSUE #968: Store in a centralized place which one is the current default model
Each model comes with its own README, which lists its supported types. So, a prospective user does have a way to check that out. However, a user does not necessarily known which one is the default model for the rust client and the python module, and she is left guessing what's really supported.

One solution is to document in the main README which model is the current default, and add a link to such model's README. The risk though is that that information becomes stale at some point.

What we could do is to have one place (e.g., `<github_repo_path>/.default-model-name`) where we indicate which one is the default model, and then we automate the checks of the docs (we could integrate this in #966), and the python/rust/etc. bindings tests to make the information does not get stale.

@ia0, thoughts?

reyammer:
Fixes #968.

### DIFF
--- a/python/scripts/check_documentation.py
+++ b/python/scripts/check_documentation.py
@@ -62,14 +62,19 @@ def check_versions_are_up_to_date() -> bool:
     # Check that the versions mentioned in the READMEs are up to date
     python_latest_stable_version = get_python_latest_stable_version()
     python_default_model_name = get_python_default_model_name()
+    rust_default_model_name = get_rust_default_model_name()
 
-    print(f"INFO: {python_latest_stable_version=} {python_default_model_name=}")
+    print(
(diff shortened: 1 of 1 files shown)


python/scripts/check_documentation.py
```diff
@@ -108,6 +113,12 @@ def get_python_default_model_name() -> str:
     return default_model_name
 
 
+def get_rust_default_model_name() -> str:
+    model_symlink_path = REPO_ROOT_DIR / "rust" / "gen" / "model"
+    assert model_symlink_path.is_symlink()
+    return model_symlink_path.readlink().name
```
ia0, line 119: You don't need to step the directory? I would expect this to output `../../assets/models/standard_xx`

ia0: I see, SG

ia0 (approved):

PR #1010, opened 2025-03-24
Upload the latest tfjs models to the website

invernizzi:


reyammer (approved):
rename like we do in assets (or something like that): https://github.com/google/magika/tree/main/assets/models/standard_v3_2

PR #1160, opened 2025-10-02
Revert "build(deps): bump commander from 13.1.0 to 14.0.1 in /js"

invernizzi:
Reverts google/magika#1141


invernizzi:
Reverting this as it breaks tests on GH Actions, but not locally

PR #1167, opened 2025-10-06
Fix rustdoc nightly build: Replace doc_auto_cfg with doc_cfg

ishanrajsingh:
Fixes #1166

## Summary
Updates the rustdoc feature flag from the deprecated `doc_auto_cfg` to `doc_cfg`.

## Background
The `doc_auto_cfg` feature was merged into `doc_cfg` in Rust nightly as part of RFC 3631 (rust-lang/rust#138907). The old feature name is no longer available and causes compilation errors.

## Changes
- Updated `src/lib.rs` line 34 to use `feature(doc_cfg)` instead of `feature(doc_auto_cfg)`

## Testing
- Verified locally by running `./test.sh` successfully
- All tests pass including doc generation

PR #1168, opened 2025-10-06
Keep ia0 as owner of rust Changelogs

ia0:
The `*.md` rule for documentation is a bit strong. The intent was only to capture READMEs under rust, so this PR explicitly excludes the CHANGELOGs from that rule. This makes it possible to do pure code changes to the Rust CLI and library without double approvals.


ia0:
See #1167 for an example PR requiring double approvals.

PR #1198, opened 2025-10-22
tests: add Swift and Python bytecode samples

dukecat0:
- Swift code was created by me manually.
- Python bytecode is generated using the following code:
```py
import py_compile
py_compile.compile('fib.py')  
```
`fib.py`:
```py
def fib(n):
    if n <= 1:
        return n
    else:
        return fib(n-1) + fib(n-2)

print(fib(10))
```
Related to #662


reyammer:
Merged, thanks!

PR #1206, opened 2025-10-29
python: mark python 3.14 as supported

### LINKED ISSUE #1277: Python 3.14 support and updated onnxruntime dep
This library's dependency on https://pypi.org/project/onnxruntime/ had been blocking more flexible pinning (esp for windows) & python 3.14 support.

It should be unblocked now as of the Feb 5th release of the onnxruntime python package.

Magika is a heavily used dep & python 3.14 support will help unblock downstream libs & end users move forward.

reyammer:
~Blocked: onnxruntime does not yet support python 3.14.~

onnxruntime 1.24.1+ does support python 3.14

Fixes #1277.

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -36,10 +36,21 @@ version = "1.0.2.dev0"
 
 dependencies = [
     "click>=8.1.7",
+
+    # Python 3.14+: Min version 1.24.1
     "onnxruntime>=1.24.1 ; python_version >= '3.14'",
-    "onnxruntime>=1.17.0 ; python_version > '3.9' and python_version < '3.14'",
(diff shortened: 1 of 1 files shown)


fmmoret:
Onnxruntime was just upgraded on Feb 5th! It's been a while 
https://github.com/google/magika/issues/1277

reyammer:
you mean the additional constraint on numpy? it seems uv had troubles selecting a proper version for it when on 3.14. It's a bit tricky because both magika and onnxruntime do depend on numpy. But I'm checking in #1268 whether we could get rid of this dependency, I don't think we really need it.

Thanks for the feedback!

PR #1223, opened 2025-11-02
docs: fix OpenSSF Scorecard README badge

### LINKED ISSUE #1061: Fix OpenSSF Scorecard badge
The badge link seems correct (and in the past it worked well), but recently the badge started having rendering issues (both on github and on pip's page).

Image

Commented out for the time being.

ashmod:
Fixes #1061


reyammer:
Thanks! Just checked again the report, and it seems there are many false positives. I guess that dealing with n bindings and complicated rust/python build processing doesn't help... So I'm merging the fix but I'll comment it out for now, I'll look into fixing those false positives before re-enabling this one. Thanks again!

reyammer (approved):

PR #1226, opened 2025-11-05
Rename the Rust library when patching the CLI version for Maturin

### LINKED ISSUE #1213: build pipeline breaks when Magika python module has the same version as magika rust library
I was trying to build 1.0.0 of the python package, but, after the version patch we do in `fix_package_version.py`, `cargo check` started complaining.

It seems that both magika rust binary and the magika client show up as two different "magika package" in `Cargo.lock`; and after we patch magika client's to be the same as the magika python module... well, if those two are the same, `cargo check` complains that there are two packages in the lock file, with the same name, and with the same version.

This is not such a big deal (I just bumped magika's python version to 1.0.1), but it's annoying (and when magika rust library goes to 1.0.1, well, that will break things).

AFAIK we need to do this weird "let's patch Cargo.toml with the python version we want" because, when building the python package, maturin uses Cargo.toml instead of pyproject.toml. But maybe things have changed in the meantime?

/cc @ia0

ia0:
Note that `fix_package_version.py` is already a hack to work around https://github.com/PyO3/maturin/issues/2163. This PR is to additionally work around https://github.com/rust-lang/cargo/issues/10353. Once the Maturin issue is fixed, all those work-arounds can be deleted.

Fixes #1213

### DIFF
--- a/python/scripts/fix_package_version.py
+++ b/python/scripts/fix_package_version.py
@@ -28,6 +28,7 @@ def main() -> None:
     # Compute paths to files we'll need to restore at the end of the build
     rust_main_rs_path = rust_root_dir / "cli" / "src" / "main.rs"
     rust_cli_cargo_toml_path = rust_root_dir / "cli" / "Cargo.toml"
+    rust_lib_cargo_toml_path = rust_root_dir / "lib" / "Cargo.toml"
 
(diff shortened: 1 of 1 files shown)


reyammer (approved):

reyammer:
That is some excellent python code, thanks!

PR #1228, opened 2025-11-08
Add basic ruby code

yosefbennywidyo:
Introduce sample ruby code.

Related with #662


reyammer (commented):
tests should be placed in `tests_data/basic/ruby/code.rb` but this seems to place them in `basic/*`?

yosefbennywidyo:
@reyammer I've moved it to the correct folder

Issue #1144, opened 2025-09-30
[NEW CONTENT TYPE REQUEST] cannot identify mobi ebook format


lihuibng:
**What type of file would you like magika to detect?**
Write the full name of the file format, followed by the file extention in parenthesis.
Examples:
 - "ebook mobi  (.mobi)"

PR #1232, opened 2025-11-09
feat: add mobi format

Ajay-Satish-01:
#1144


invernizzi:
Hi Ajay, this merge request doesn't actually introduce `.mobi` support. The model would need to be trained with `.mobi` files for that to happen (about 10k, and with no licensing issues). Most of the files that are being edited manually here are automatically generated as a result of that training.
I'll close this pull request for now.

Ajay-Satish-01:
> Hi Ajay, this merge request doesn't actually introduce `.mobi` support. The model would need to be trained with `.mobi` files for that to happen (about 10k, and with no licensing issues). Most of the files that are being edited manually here are automatically generated as a result of that training. I'll close this pull request for now.

I was just wondering about it. Thanks for confirming it as well.

PR #1240, opened 2025-11-12
Fix typos

kianmeng:
Found via `codespell -S *.json,*.lock,tests_data -L hve,afterall` and `typos --hidden --format brief`

### DIFF
--- a/python/scripts/test_magika_model.py
+++ b/python/scripts/test_magika_model.py
@@ -59,12 +59,12 @@ def main(model_dir_or_name: str) -> None:
             mr = m.identify_path(test_path)
             assert mr.ok
 
-            predicated_content_type = mr.prediction.output.label
+            predicted_content_type = mr.prediction.output.label
             expected_content_type = test_path.parent.name
-            if predicated_content_type != expected_content_type:
+            if predicted_content_type != expected_content_type:
                 with_error = True
                 log_error(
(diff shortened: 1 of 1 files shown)


reyammer:
Thanks! This touches many files, and I'll first need to update the versioning of the packages as well. I'll get to it.

BTW, is the git diff coming from just using those tools above automatically, or you used the tools and then you manually checked what were reasonable findings? I'm wondering whether we could add these commands as part of our CI...

kianmeng:
> Thanks! This touches many files, and I'll first need to update the versioning of the packages as well. I'll get to it.

Noted.

> BTW, is the git diff coming from just using those tools above automatically, or you used the tools and then you manually checked what were reasonable findings? I'm wondering whether we could add these commands as part of our CI...

Manually check one by one, as suggested by the tools. Corrections were applied based on my understanding of the context or intended usage of the doc/code.

kianmeng:
> See my other comment's on gemini's feedback. Thanks!

@reyammer Done. See https://github.com/google/magika/pull/1240/changes/70164563efcc8bcbfbf2a285f21449bf8b6fa651 and https://github.com/google/magika/pull/1240/changes/4dceefeeffc3cdb2ad233b77a467913744c664fc

reyammer:
Merged, thanks so much!

kianmeng:
:partying_face: :partying_face: :partying_face: :partying_face: :partying_face:

Issue #1248, opened 2025-11-20
[Misdetection] .env file misdetected as HTML


petrkotas:
**What should the file have been detected as? What has the file been misdetected as?**
I have tried to input the .env file with edge case content

```
text="<html><head></head><body></body></html>"
value="1234"
body="<h1>kokot</h1>"
```

it got classified as HTML. 

<img width="969" height="755" alt="Image" src="https://github.com/user-attachments/assets/5be71409-b34c-4994-8a14-a2fb2a7f6163" />

It should be .env, but it is clearly and edge case. 

**Please link or attach the misdetected file below** (Do NOT upload PII!)
File included up.

**Additional context**
None

PR #1252, opened 2025-11-27
Add basic .env file

yosefbennywidyo:
Introduce sample .env file

Related with #662 and #1248


reyammer:
env is not a supported content type. This specific sample is (correctly?) flagged as txt, but at the moment we are not storing samples in the tests for generic txt.

Issue #1249, opened 2025-11-20
[Misdetection] texfile file misdetected as ignorefile


petrkotas:
**What should the file have been detected as? What has the file been misdetected as?**
Tried blabla text file, got detected as ignorefile

```
klsdjcdksaljckldsaj
sdakchdsajkchds
asjkdchdsajkh
```

I understand that this is an edge case. 

<img width="969" height="755" alt="Image" src="https://github.com/user-attachments/assets/1355ba13-5264-4922-9803-b071d4025c73" />

**Please link or attach the misdetected file below** (Do NOT upload PII!)
Provided up

**Additional context**
None

invernizzi:
Thanks! I can see why magika detects it as an ignore file. It's probably not what we want as those lines do not look like filenames, but yes it's an edge case. Perhaps we can augment the dataset to make this distinction.

PR #1253, opened 2025-11-27
Add other ignorefile sample

yosefbennywidyo:
related to: #662 and #1249


reyammer (approved):

reyammer:
Thanks for the PR! Note for the future: at this stage we are mostly interested in covering content types that are not currently covered by the basic tests, rather than small variations of types we already have something for. This one for example is borderline similar to the existing one.

PR #1257, opened 2025-12-11
Add internal _time feature to the Rust CLI

ia0:


reyammer:
Was doing some clean up and saw this, makes sense to merge this one, seems useful?

ia0:
> Was doing some clean up and saw this, makes sense to merge this one, seems useful?

It depends how often we need to do measure performance. From experience it doesn't seem that often, so I would say we simply close this.

ia0:
Closing as obsolete. We have a `_trace` feature for a similar purpose (which could be extended).

PR #1263, opened 2026-01-09
python: fix pre-release checker script

reyammer:
Github released new runners 3 days ago (https://github.com/actions/runner-images/releases/tag/ubuntu24%2F20260105.202), and since then the pre-release check script started failing. (There is  a change the timing is just an unlucky coincidence.)

Root cause and fix: we try to get the rust's magika CLI version via `magika -V`, but now this errors out with "magika not found". Since we install magika via `uv`, this is actually expected, and it's not clear how it worked in the past. Now doing `uv run magika -V` instead.

Note also that we cannot always run magika with `uv run magika`: this seems to break the pre-release script when running it after installing magika via pip on Windows. For this reason, we actually add a new CLI option `--run-magika-via-uv` to control how we try to run the magika client.


reyammer:
This PR is now not needed anymore. We tracked down the problem: the regression is due to a new version of maturin (from Jan 5th), which is actually the only dependency that we did not have pinned. Pinned maturin in https://github.com/google/magika/pull/1264 as a workaround, we will check with the maturin folks to see whether it's a bug or we are doing something wrong.

PR #1296, opened 2026-02-19
chore: update ort and ndarray

cr-EliasJunior:
This PR updates the ort dependency from 2.0.0-rc.10 to 2.0.0-rc.11, along with the corresponding ndarray version update required by the new release.

## Background

When running on Windows Server, the ONNX Runtime (ort) attempts to load DLLs that are only available on Windows Desktop SKUs (e.g. WLANAPI.DLL, DWM-related API sets, and others). On Windows Server these DLLs are absent, causing a hard panic inside ort during initialization with no opportunity for the caller to handle the failure gracefully.
ort 2.0.0-rc.11 includes a fix ([pykeio/ort@8b3a1ed](https://github.com/pykeio/ort/commit/8b3a1edd1ca4e10e8f6c16e05091f0e17cd9da23)) that allows consumers to detect and handle this initialization failure gracefully instead of panicking, enabling callers to disable ONNX-dependent functionality at runtime rather than crashing the process.

## Changes

Bumped ort to =2.0.0-rc.11
Updated ndarray to 0.17 to match ort's updated requirement
Adjusted Tensor::from_array call sites to conform to the updated OwnedTensorArrayData trait bounds in rc.11

### DIFF
--- a/rust/lib/Cargo.lock
+++ b/rust/lib/Cargo.lock
@@ -45,25 +45,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
-name = "base64ct"
-version = "1.8.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


reyammer:
Thanks! It seems the CLA check failed, would you mind looking into that so that we can accept your contribution? Thanks!

era:
Hi @reyammer, let me use my personal account instead: https://github.com/google/magika/pull/1297

PR #1312, opened 2026-03-01
rust: pin ort/ndarray for x86_64 build compatibility

chenrui333:
## Summary
- Pin Rust `ort`/`ort-sys` to `2.0.0-rc.10`
- Align `rust/lib` `ndarray` to `0.16.1`
- Update lockfiles for `rust/cli` and `rust/lib`

## Context
- Fixes x86_64 build breakage seen in Homebrew core bump PR
- Ref: https://github.com/Homebrew/homebrew-core/pull/269338

### DIFF
--- a/rust/cli/Cargo.lock
+++ b/rust/cli/Cargo.lock
@@ -125,6 +125,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2261d10cca569e4643e526d8dc2e62e433cc8aba21ab764233731f8d369bf394"
 
 [[package]]
+name = "block-buffer"
+version = "0.10.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3078c7629b62d3f0439517fa394996acacc5cbc91c5a20d8c658e77abd503a71"
+dependencies = [
+ "generic-array",
+]
+
+[[package]]
 name = "byteorder"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


ia0:
> * Fixes x86_64 build breakage seen in Homebrew core bump PR

The support for x86_64-apple-darwin has been removed, see #1302 (we can't test this target on Github). Version 1.0.1 should be used for this target. Is this something possible to do in Homebrew?

chenrui333:
> > * Fixes x86_64 build breakage seen in Homebrew core bump PR
> 
> The support for x86_64-apple-darwin has been removed, see #1302 (we can't test this target on Github). Version 1.0.1 should be used for this target. Is this something possible to do in Homebrew?

does that mean x86_64 is no longer supported?

ia0:
> does that mean x86_64 is no longer supported?

Do you have homebrew download statistics sliced by architecture (for magika and in general)? It would be useful to know how wide spread this architecture (which doesn't have a GitHub runner) still is.

chenrui333:
> > does that mean x86_64 is no longer supported?
> 
> Do you have homebrew download statistics sliced by architecture (for magika and in general)? It would be useful to know how wide spread this architecture (which doesn't have a GitHub runner) still is.

yeah, we dont have specific architecture usage report for specific formula. But if no support, we can drop on our side.

ia0:
> yeah, we dont have specific architecture usage report for specific formula. But if no support, we can drop on our side.

Thanks! Yes then please drop the support on your side. I suspect it's used by less than 5% users and will continue to get less and less used over time.

chenrui333:
> > yeah, we dont have specific architecture usage report for specific formula. But if no support, we can drop on our side.
> 
> Thanks! Yes then please drop the support on your side. I suspect it's used by less than 5% users and will continue to get less and less used over time.

Got you, that works. 👍

chenrui333:
@ia0 removing the patch also broke the linux arm build, raise a separate issue, https://github.com/google/magika/issues/1323

you can see the detailed build log in here, https://github.com/Homebrew/homebrew-core/actions/runs/23243289467/job/67573224488?pr=272963

PR #1313, opened 2026-03-02
rust: Add identify_reader_sync/async for Read + Seek types

withbitsinmind:
This PR adds two new functions to session, `identify_reader_sync` and `identify_reader_async` which makes it possible to pass in any type that implements `Read` and `Seek`. It is analogous to the `identify_stream` function in the Python implementation #979.

I suppose an alternative, and perhaps cleaner solution, would be to drop the sealed trait pattern for `SyncInput` and `SyncInputApi` (and the async equivalent) and instead expose the `SyncInputApi` so that users can implement it for arbitrary types.

### DIFF
--- a/rust/cli/publish.sh
+++ b/rust/cli/publish.sh
@@ -16,7 +16,7 @@
 set -e
 . ../color.sh
 
-# This script rename the package to magika-cli to publish to crates.io.
+# This script renames the package to magika-cli and publishes to crates.io.
 
 [ -z "$(git status -s)" ] || error "Repository is not clean"
 
@@ -34,4 +34,11 @@ cargo fmt -- src/main.rs
 info "Run tests"
 ./test.sh
 
-todo "Create a temporary commit and run cargo publish"
+info "Create a temporary commit"
+git commit -aqm'Temporary commit to publish magika-cli'
+
+info "Publish to crates.io"
+cargo publish
+
+info "Revert the temporary commit"
+git reset --hard HEAD~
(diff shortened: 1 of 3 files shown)


ia0:
Thanks for the PR!

Do you have an example of an object that implements `Read` and `Seek` and is not `File`? This could help figure out the best solution.

I see at least the following options:
- Add the `identify_reader_{sync,async}` functions as you did. I don't think we should do that. We should use the `identify_content_{sync,async}` functions directly.
- Add the `ReadSeek` wrapper as you did. This doesn't look great.
- Unseal `{Sync,Async}Input` as you proposed. I guess that's an option, but then we need to decide if it should be an unsafe trait.
- Provide a blanket implementation for `Read` and `Seek`. We need to make sure this is not a breaking change.

withbitsinmind:
The example object in my case is `HttpRangeReader`, a private struct, which implements `Read` and `Seek` (or the async equivalents) using HTTP Range requests towards a URL. This implementation would make it possible to run magika on said resource without having to download it to a file beforehand. It could for example be the presigned URL for a large S3 object.

As for the unseal alternative, my take is that it should not be marked as unsafe because it is not in itself unsafe, though an implementor could of course provide an incorrect or unsafe implementation.

ia0:
> The example object in my case is `HttpRangeReader`, a private struct, which implements `Read` and `Seek` (or the async equivalents) using HTTP Range requests towards a URL.

Thanks! This makes sense. Could you tell me if #1315 would work for you?

withbitsinmind:
Brilliant! I can confirm that #1315 works for me 👍

PR #1318, opened 2026-03-11
feat: add Magika.get_content_type_info() public API method (fixes #826)

### LINKED ISSUE #826: Expose content type knowledge base information
In a previous version of the magika python package (e.g., `0.5.1`), we had a `ContentTypesManager` class that was wrapping the knowledge about content types and their metadata (e.g., mime types). This class was never meant to be used by external clients (nor it was documented). In the new version, this class has been refactored away, now content type labels are actual enums, and we track their metadata via a private field in the `Magika` object.

That being said, the content types knowledge base could be of some use to external clients. It may be worth exploring a clean way to expose this information.

One idea is to add a couple of methods to the `Magika` object to list all content types (all in our content types KB, or all supported by a given model) and one method to access the metadata (e.g., mime type) for a given content type.

shivamtiwari3:
## Summary

Fixes #826.

The content type knowledge base (mime type, group, description, extensions, is_text) was only accessible via the private `_get_ct_info()` method and the internal `_cts_infos` dict. This PR exposes a public `get_content_type_info(label)` method, completing the content-type introspection API alongside the existing `get_output_content_types()` and `get_model_content_types()` methods.

---

## Root Cause

`Magika._get_ct_info()` at `magika.py:344` held all the content type metadata but was private and undocumented. External clients had no supported way to query mime type, group, description, or file extensions for a given `ContentTypeLabel`.

---

## Solution

Added `Magika.get_content_type_info(label: ContentTypeLabel) -> ContentTypeInfo` in `python/src/magika/magika.py`. The method:
- Returns the `ContentTypeInfo` for any label present in the knowledge base
- Raises `MagikaError` (rather than a bare `KeyError`) if the label is not in the KB, giving callers a clear error message
- Is a thin, clean wrapper around the existing `_cts_infos` dict — no new state introduced

The method fits naturally alongside `get_output_content_types()` and `get_model_content_types()`: those methods answer *"which labels exist?"* while `get_content_type_info()` answers *"what does this label mean?"*

---

## Testing

- Added `test_get_content_type_info` in `python/tests/test_magika_python_module.py` that:
  - Calls `get_content_type_info()` for every label returned by `get_output_content_types()` and asserts all fields are valid
  - Spot-checks `ContentTypeLabel.PDF` for exact expected values (`application/pdf`, group `document`, `is_text=False`)
  - Verifies `ContentTypeLabel.PYTHON` returns `is_text=True`
- All 33 existing tests pass: `uv run pytest tests -m "not slow"`
- `ruff check`, `ruff format`, and `mypy` pass clean

---

## Checklist

- [x] Fixes the root cause (private method exposed via clean public API)
- [x] New test covers all output content types and spot-checks specific metadata
- [x] All existing tests pass
- [x] No unrelated changes
- [x] Code style matches project conventions
- [x] Read CONTRIBUTING.md

### DIFF
--- a/python/src/magika/magika.py
+++ b/python/src/magika/magika.py
@@ -272,13 +272,14 @@ class Magika:
 
         Raises:
             MagikaError: If the label is not present in the content type
-                knowledge base (e.g., ContentTypeLabel.UNDEFINED).
+                knowledge base.
         """
-        if label not in self._cts_infos:
+        try:
+            return self._cts_infos[label]
+        except KeyError:
             raise MagikaError(
                 f"Content type '{label}' is not in the content type knowledge base."
-            )
-        return self._cts_infos[label]
+            ) from None
 
     @staticmethod
     def _get_default_model_name() -> str:
--- a/python/src/magika/magika.py
+++ b/python/src/magika/magika.py
@@ -256,6 +256,30 @@ class Magika:
         model_content_types.update(self._model_config.target_labels_space)
         return sorted(model_content_types)
 
+    def get_content_type_info(self, label: ContentTypeLabel) -> ContentTypeInfo:
+        """Returns metadata for a given content type label.
+
+        This provides access to the content type knowledge base, exposing
+        structured metadata such as mime type, group, human-readable
+        description, common file extensions, and whether the type is
+        text-based.
+
+        Args:
+            label: The ContentTypeLabel to look up.
+
+        Returns:
+            The ContentTypeInfo for the given label.
+
+        Raises:
+            MagikaError: If the label is not present in the content type
+                knowledge base (e.g., ContentTypeLabel.UNDEFINED).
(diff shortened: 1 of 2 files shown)


shivamtiwari3:
Applied all three suggestions:
1. **EAFP pattern** — switched from `if label not in self._cts_infos` to `try/except KeyError` with `raise ... from None` to suppress the chained exception context.
2. **Docstring fix** — removed the incorrect `ContentTypeLabel.UNDEFINED` example from the `Raises` section (it is present in the KB and will not raise).
3. **Error test** — added `pytest.raises(MagikaError, ...)` assertion at the end of `test_get_content_type_info` using a non-existent label string.

shivamtiwari3:
Hi — happy to make any further changes if helpful. Just checking if this is still under consideration.

PR #1335, opened 2026-03-23
Fix: feature extraction nullification via whitespace-padded inputs leading to security gate bypass

ParzivalHack:
`_extract_features_from_seekable()` reads at most `block_size` (4096) bytes from offset 0 and calls `lstrip()` on the result. If a file starts with more than 4096 bytes of whitespace, the read window contains only whitespace, `lstrip()` returns `b""`, and the model receives an all `padding_token` input (misclassifying any file as `txt`, regardless of actual file content).

The existing fallback check in `_get_result_or_features_from_seekable()`, detects the all-padding condition, but reads from offset 0 unconditionally, retrieving the same attacker-controlled whitespace. `_get_label_from_few_bytes()` then decodes that whitespace as valid UTF-8 and returns `ContentTypeLabel.TXT` unconditionally.

This is a security issue i reported to Google's OSS VRP, which explicitly told me to open a public PR first. I already tested the patch locally and it works on my end, so let me know :)

### DIFF
--- a/python/src/magika/magika.py
+++ b/python/src/magika/magika.py
@@ -436,16 +436,8 @@ class Magika:
         assert end_size < block_size
         assert not use_inputs_at_offsets
 
-        # we read at most block_size bytes
-        bytes_num_to_read = min(block_size, seekable.size)
-
         if beg_size > 0:
-            # Read at most `block_size` bytes from the beginning; `lstrip()``
-            # them (or `strip()` them if the file size is less or equal than
-            # `block_size`); take at most `beg_size` bytes, and optionally pad
-            # them with `padding_token` to get to a list of `beg_size` integers.
-            beg_content = seekable.read_at(0, bytes_num_to_read)
-            beg_content = beg_content.lstrip()
+            beg_content = Magika._read_beg_content(seekable, block_size)
             beg_ints = Magika._get_beg_ints_with_padding(
                 beg_content, beg_size, padding_token
             )
@@ -453,15 +445,7 @@ class Magika:
             beg_ints = []
 
         if end_size > 0:
-            # Read at most `block_size` bytes from the end; `rstrip()`` them (or
-            # `strip()` them if the file size is less or equal than
-            # `block_size`); take at most `end_size` bytes (from the end), and
-            # optionally pad them (at the beginning) with `padding_token` to get
-            # to a list of `end_size` integers.
-            end_content = seekable.read_at(
-                seekable.size - bytes_num_to_read, bytes_num_to_read
-            )
-            end_content = end_content.rstrip()
+            end_content = Magika._read_end_content(seekable, block_size)
             end_ints = Magika._get_end_ints_with_padding(
                 end_content, end_size, padding_token
             )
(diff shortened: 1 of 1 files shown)


ParzivalHack:
Just signed the necessary CLA as well (with this github username) :)

ParzivalHack:
Hi @reyammer, mind checking this out when you can? Thanks!

reyammer:
Hello, sorry for the delay in getting back to you!

Thanks for reporting.

A couple of questions, just to make sure I got it right:
- About the bug: you would consider this as a security vulnerability, but under which angle? Is it because one could trick magika to misdetect a file of type X as a file of type Y?
- About the patch: my understanding is that if the first block is full of whitespace, the proposal would be to read additional blocks until something is found (or the file is over)?

ParzivalHack:
> Hello, sorry for the delay in getting back to you!
> 
> Thanks for reporting.
> 
> A couple of questions, just to make sure I got it right:
> 
> * About the bug: you would consider this as a security vulnerability, but under which angle? Is it because one could trick magika to misdetect a file of type X as a file of type Y?
> * About the patch: my understanding is that if the first block is full of whitespace, the proposal would be to read additional blocks until something is found (or the file is over)?

Hey @reyammer, no worries, thanks for getting back to me!

About the security bug: yes, exactly, the security angle is that an attacker can trick Magika into misdetecting a malicious file of type X as `txt` or `unknown`. The concern is specifically for deployments that use Magika as a security gate or pre-filter (i.e. blocking uploads of active content like PHP, JavaScript, or Python, based on Magika's classification). By prepending more than 4096 bytes of whitespace to any payload, the classification always falls back to `txt`, allowing the file to bypass the gate. The padded file remains fully valid and executable, since PHP, Python, Node.js, and browsers all ignore leading whitespace. And about the patch, yeah, my proposal is to scan forward in `block_size` chunks, until non-whitespace bytes are found, rather then reading a fixed block from offset 0 and getting an empty buffer after stripping. Same thing (but applied in reverse) for the end buffer, and also applied to the fallback read path in `_get_result_or_features_from_seekable()`, which has the same issue.

Thanks again :)

reyammer:
Ok thanks for the clarification and thanks again for the PR. Unfortunately this is a known limitation of magika's current version, and it goes well beyond dealing with whitespaces. Especially for textual content types, one could add something like N bytes of comments in language X, and then switch to language Y for the rest of the file, which is not visible to the model at all. It's a bit trickier to bypass so easily with binary content types, but it's definitively possible with some fuzzing-based and other approaches. In essence, using magika as the only gatekeeper for security related situations is not advisable; the security-related use cases we are OK reccomending is in scenarios where "it's never worse" to use magika than to not use it (e.g., scan PDF-looking files with PDF-focused scanners, and use magika in addition to existing content type detections tools, which may miss something).

One note about the PR: magika intentionally only checks up to 4096 bytes because for serious prod deployments we want/need near constant-time inference speed; if we start reading N blocks, the worst case inference speed could become linear wrt the size of the file. Since scanning more bytes doesn't fundamentally fix the underlying issue, we decided to keep the constant upper bound inference time, which is a very good property.

Last note: if one wants to tweak about how the file is read, remove more whitespaces, implement a sliding-window approach and things like this, then the python library should make it trivial to implement such variants.

I hope the rationale is clear, and thanks again!

PR #1341, opened 2026-04-15
Refactor formatPackage for directory creation and overrides

krataratha:
Enhance formatPackage function to create output directories and allow extra fields.


reyammer:
Hello, thanks for the PR. Due to time constraints, at this point in time we are not accepting contributions/edits on internal parts that are not motivated by actual use needs (but that would require extra testing on our end). Sorry, and thanks again.

PR #1342, opened 2026-04-15
fix(magika): prevent whitespace-prefix bypass in fallback detection path

herdiyana256:
When a file has >= block_size bytes of leading ASCII whitespace followed by a binary payload, the existing fallback path misclassified the file as 'txt'. The root cause was a two-step failure:

1. _extract_features_from_seekable reads block_size (4096) bytes and calls lstrip(), producing an empty buffer if all bytes are whitespace. This fills beg[] with all padding_token values.

2. _get_result_or_features_from_seekable detects the all-padding condition and falls back to _get_result_from_few_bytes. However, it re-read the *same* first block_size bytes (still all whitespace), passed them to _get_label_from_few_bytes, which only checks UTF-8 validity. Whitespace is valid UTF-8, so the method returned ContentTypeLabel.TXT — ignoring the actual payload hidden beyond byte 4096.

Fix: In the fallback path, strip the leading whitespace from the first block. If the result is empty *and* the seekable has content beyond block_size, advance past the whitespace prefix and read the next block (which contains the actual payload) before calling _get_result_from_few_bytes.

This preserves the existing behaviour for:
- Legitimate files that are shorter than block_size (no change).
- Files with *some* leading whitespace that does not fill the entire first block (stripped_first_block is non-empty; existing path taken).
- Files whose entire content is whitespace (seekable.size <= block_size; the bypass condition is not triggered).

Only files where ALL of the first block_size bytes are whitespace AND more content follows are now routed to the new payload-reading path.

Regression test added:
test_magika_module_whitespace_prefix_does_not_bypass_binary_detection verifies ELF, PE/MZ, and non-UTF-8 binary payloads are not classified as TXT when prepended with block_size whitespace bytes.

### DIFF
--- a/python/src/magika/magika.py
+++ b/python/src/magika/magika.py
@@ -757,10 +757,41 @@ class Magika:
                 == self._model_config.padding_token
             ):
                 # If the n-th token is padding, then it means that,
-                # post-stripping, we do not have enough meaningful
-                # bytes.
-                bytes_to_read = min(seekable.size, self._model_config.block_size)
-                content = seekable.read_at(0, bytes_to_read)
+                # post-stripping, we do not have enough meaningful bytes
+                # in the first block.
+                #
+                # Security note: this path is also reachable when an attacker
+                # prepends ≥block_size bytes of ASCII whitespace to a binary
+                # payload (e.g. ELF, PE, PHP webshell).  In that scenario,
+                # re-reading the same first block (all whitespace) and feeding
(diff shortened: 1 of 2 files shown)


reyammer:
Hello, thanks for the PR. Unfortunately this approach doesn't solve the problem at its root and introduces additional problems. More context here: https://github.com/google/magika/pull/1335#issuecomment-4197976790

Issue #1343, opened 2026-04-15
[Misdetection/Security] PHP file misdetected as txt due to low-confidence fallback


CheonWoong-Park:
I initially reported this to the Google OSS VRP (https://g.co/vulnz).
They acknowledged it as a valid product vulnerability. However, because Magika is classified as an OT2 project, they require a patch to be committed and merged before they can process the report.
They encouraged me to open a public issue to collaborate with the maintainers on a fix, so I am doing that here.

Magika extracts features only from the beginning and end of a file (`beg_size=1024`, `end_size=1024`). For files smaller than 1024 bytes, both feature windows contain the exact same content.

Knowing that the goal was simply to keep the score below 0.5, I used Claude to help craft a PoC.

In the PoC below, the file is 958 bytes long: roughly 928 bytes of HTML/JS followed by a 30-byte `<?php ... ?>` block. Because the HTML/JS content dominates both feature vectors, the PHP signal is relatively weak.

As a result, the model predicts `dl=php`, but with a score of about 0.489. Due to the low-confidence fallback in `python/src/magika/magika.py`, the final result becomes `output=txt`. In other words, Magika returns `txt` for a file that it internally considers likely to be PHP.

This is also highly position-sensitive. If the PHP block is moved slightly, the score fluctuates across the threshold (for example, 0.506, 0.489, and 0.646). This appears consistent with the current feature extraction design and with where the PHP bytes land inside the two feature windows.

The attached PoC reproduces this behavior and shows that the PHP interpreter still successfully executes the sample.

Because this is my first time handling a security issue through a public issue tracker, I was not sure whether including the full PoC here would be appropriate, so I have not attached it for now.

Nick-Bug-Bounty-Hunter:
Hi maintainers — I took a quick look at this and it seems plausible to address with a targeted change around the low-confidence fallback / small-file handling, plus a regression test. If that direction matches what you would want, I can prepare a draft PR rather than guessing at the preferred fix shape.

PR #1344, opened 2026-04-15
fix: whitespace-prefix binary bypass — check both beg and end features

JasonOA888:
## Bug

Binary files prefixed with `>= block_size` whitespace bytes were always classified as **TXT**, regardless of actual content.

### Root Cause

The post-strip validation in feature extraction only checked `beg` features. After `lstrip()` removes the whitespace prefix, `beg` becomes all padding tokens, triggering a fallback to a simple UTF-8 check on the raw first block. Since whitespace is valid UTF-8, the result is always TXT.

The `end` features (extracted from the tail of the file) **do contain real content**, but were never consulted.

### Reproduction

```python
from magika import Magika
m = Magika()

# Binary ELF with whitespace prefix
elf = b" " * 4096 + b"\x7fELF\x02\x01\x01\x00" + b"\x00" * 500
res = m.identify_bytes(elf)
print(res.output.label)  # Before: txt (wrong) | After: unknown (correct)

# Pure whitespace still works correctly
res = m.identify_bytes(b" " * 5000)
print(res.output.label)  # txt ✓
```

### Fix

Check **both** `beg` and `end` features. Only fall back to UTF-8 check when **both** are all padding after stripping. Applied consistently to:

- **Python** (`python/src/magika/magika.py`)
- **Rust** (`rust/lib/src/input.rs`)
- **JS** (`js/magika.ts`)

### Additional fix (JS only)

The JS low-confidence fallback checked `model_prediction.label.is_text` (pre-overwrite) instead of `output_label.is_text` (post-overwrite), inconsistent with Python. This is a latent bug — with the current overwrite_map (`randomtxt→txt`, `randombytes→unknown`) it produces identical results since both sides have the same `is_text` value, but it would diverge with future overwrite_map entries.

Related: #1343, #1342

### DIFF
--- a/js/magika.ts
+++ b/js/magika.ts
@@ -202,7 +202,7 @@ export class Magika {
         this.model_config.medium_confidence_threshold)
     ) {
       overwrite_reason = OverwriteReason.LOW_CONFIDENCE;
-      if (this.cts_infos[model_prediction.label].is_text) {
+      if (this.cts_infos[output_label].is_text) {
         output_label = ContentTypeLabel.TXT;
       } else {
         output_label = ContentTypeLabel.UNKNOWN;
--- a/python/src/magika/magika.py
+++ b/python/src/magika/magika.py
@@ -751,22 +751,28 @@ class Magika:
                 self._model_config.use_inputs_at_offsets,
             )
             # Check whether we have enough bytes for a meaningful
-            # detection, and not just padding.
-            if (
+            # detection, and not just padding. We check both beg and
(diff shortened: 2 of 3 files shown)


reyammer:
Hello, thanks for the PR. Unfortunately this doesn't solve the problem at its root, and it adds unnecessary complexity that would not benefit virtually all realistic scenarios. More context here: https://github.com/google/magika/pull/1335#issuecomment-4197976790

PR #1350, opened 2026-04-16
Added auto-detection for small .env and JSON files

nextjohnp7:
I added a simple heuristic check to identify .env and JSON patterns in very small files. This serves as a fallback for cases where the AI model might categorize small files too generically.

### DIFF
--- a/python/src/magika/magika.py
+++ b/python/src/magika/magika.py
@@ -24,6 +24,7 @@ import logging
 import os
 import time
 from pathlib import Path
+import re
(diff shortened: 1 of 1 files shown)


nextjohnp7:
@google-cla-bot check

reyammer:
Hello, thanks for the PR. However we made the decision to not use simple heuristics like this as they are too error prone.

nextjohnp7:
Oh wow, thanks for responding!
This is my first time trying this on GitHub, bro. Thanks for the feedback, I really appreciate it.

PR #1352, opened 2026-04-17
python: detect armored PGP public keys on .asc paths

### LINKED ISSUE #1338: [Misdetection] ASCII-armored GPG key file misdetected as PEM certificate
**What should the file have been detected as? What has the file been misdetected as?**
Should be something like "ASCII-armored GPG file".

This was the output:
```
$ magika mullvad-code-signing.asc 
mullvad-code-signing.asc: PEM certificate (application)
```

**Please link or attach the misdetected file below** (Do NOT upload PII!)
https://mullvad.net/media/mullvad-code-signing.asc

**Additional context**

Bortlesboat:
Closes #1338.

## Summary
- short-circuit `.asc` files that start with `-----BEGIN PGP PUBLIC KEY BLOCK-----` to `pgp`
- include that path-based result in `get_output_content_types()`
- add regression coverage for the `.asc` path case

## Testing
- `uv run -- pytest tests/test_magika_python_module.py -k "ascii_armored_pgp_key_path or output_content_types_include_path_based_pgp"`
- `uv run -- pytest tests/test_magika_python_module.py -k "not with_symlink and not with_permission_error and not with_big_file"`

### DIFF
--- a/python/src/magika/magika.py
+++ b/python/src/magika/magika.py
@@ -226,6 +226,7 @@ class Magika:
         output_content_types: Set[ContentTypeLabel] = {
             ContentTypeLabel.DIRECTORY,
             ContentTypeLabel.EMPTY,
+            ContentTypeLabel.PGP,
             ContentTypeLabel.SYMLINK,
             ContentTypeLabel.TXT,
             ContentTypeLabel.UNKNOWN,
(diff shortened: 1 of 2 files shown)


reyammer:
Hello, thanks for the PR. However we made the decision to not use simple heuristics like this as they are too error prone.

PR #1355, opened 2026-04-17
fix: handle broken pipe gracefully in Rust CLI

### LINKED ISSUE #1354: Rust CLI panics with Broken pipe when piped to head
### Summary

When the Rust CLI is piped to a consumer that closes stdout early, such as `head`, `magika` panics with `failed printing to stdout: Broken pipe (os error 32)` instead of exiting gracefully.

### Reproduction

The README currently shows this kind of usage pattern:

```shell
magika -r * | head
```

Running that reproduces the issue:

```shell
RUST_BACKTRACE=1 magika -r * | head
```

Example output:

```text
thread 'main' panicked at library/std/src/io/stdio.rs:1165:9:
failed printing to stdout: Broken pipe (os error 32)
stack backtrace:
...
9: std::io::stdio::_print
10: magika::main::{{closure}}
```

I also checked the behavior difference locally:

- `magika -r tests_data/basic >/dev/null` exits normally
- `magika -r tests_data/basic | head >/dev/null` reproduces the panic

So this appears to be limited to stdout writes after the downstream pipe has already been closed.

### Expected behavior

`magika` should treat `stdout` `BrokenPipe` as a normal pipe-termination case and exit without printing a Rust panic/backtrace.

### Actual behavior

The CLI panics from `std::io::stdio::_print`, which makes a normal shell pipeline look like a crash.

### Possible implementation direction

A narrow fix would be to special-case `io::ErrorKind::BrokenPipe` on stdout writes only, while still surfacing unrelated I/O errors and real internal panics.

From a quick look, the relevant output path seems to be in `rust/cli/src/main.rs`, where the CLI uses `print!` / `println!` for output.

### Environment

Observed on macOS (Apple Silicon) with the Rust CLI.

zouyonghe:
Closes #1354.

When the Rust CLI is used in a normal shell pipeline and the downstream consumer exits early, `magika` currently panics instead of terminating cleanly. That makes commands such as `magika -r … | head` look like crashes, even though the only failure is that stdout is no longer available.

The root cause is that the CLI writes user-facing output through `print!` and `println!`. Those macros go through `std::io::stdio::_print`, which panics on write failures, including `io::ErrorKind::BrokenPipe`. As a result, a downstream pipe closure is treated like an internal crash rather than a normal pipe-termination condition.

This change keeps the fix narrow. The CLI now writes to stdout explicitly so write failures are returned as ordinary `io::Error` values, and the top-level entrypoint treats `BrokenPipe` as a successful exit. Other errors still use the existing error-reporting path, and the per-file error exit behavior is preserved.

I also added a regression test that spawns the CLI, reads one line of output, closes the stdout pipe early, and asserts that the process exits successfully without panic output on stderr. That locks in the exact pipeline behavior reported in the issue without depending on external shell tools.

Validation:
- `cargo test exits_cleanly_when_stdout_pipe_closes_early --test broken_pipe -- --exact`
- `../target/release/magika -r ../../tests_data/basic | head >/dev/null` returned exit code 0 with `set -o pipefail`
- `rust/cli/test.sh` passed `cargo check`, `cargo build --release`, `cargo fmt -- --check`, and `cargo clippy -- --deny=warnings`; its final `/etc/shadow` assertion is Linux-specific and does not hold on macOS because `/etc/shadow` is absent there

### DIFF
--- a/rust/cli/src/main.rs
+++ b/rust/cli/src/main.rs
@@ -15,8 +15,9 @@
 use std::borrow::Cow;
 use std::collections::HashMap;
 use std::fmt::Write;
-use std::io::ErrorKind;
+use std::io::{self, ErrorKind, Write as _};
 use std::path::{Path, PathBuf};
+use std::process::{ExitCode, Termination};
 use std::sync::Arc;
 
 use anyhow::{bail, ensure, Result};
@@ -157,7 +158,15 @@ struct Experimental {
 }
 
 #[tokio::main]
-async fn main() -> Result<()> {
+async fn main() -> ExitCode {
+    match try_main().await {
+        Ok(code) => code,
+        Err(error) if is_broken_pipe(&error) => ExitCode::SUCCESS,
+        Err(error) => Result::<(), anyhow::Error>::Err(error).report(),
+    }
+}
+
+async fn try_main() -> Result<ExitCode> {
     let mut flags = Flags::parse();
     ensure!(0 < flags.experimental.batch_size, "--batch-size cannot be zero");
     // If --num-tasks is set, we don't do any guessing.
@@ -209,7 +218,7 @@ async fn main() -> Result<()> {
     }
     drop(result_sender);
     if flags.format.json {
-        print!("[");
+        write_stdout("[")?;
     }
     let mut reorder = Reorder::default();
     let mut errors = false;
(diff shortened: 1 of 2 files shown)


ia0:
Thanks for the PR!

For some reason there's a segfault. While I look into it, @reyammer what's your opinion in the behavior of `magika` when stdout is closed (like with `magika | head -n1`)? Currently `magika` fails.

ia0:
Ok, I've figured out the segfault and fixed both this issue and the segfault in #1360.

PR #1359, opened 2026-04-17
Remove the crate naming hack for Maturin

ia0:
It seems that Maturin now does the correct thing: use the Python name for the Python package. So we can rename the CLI crate to `magika-cli` (still keeping the binary name to `magika`).

With this cleanup it is now possible to have the Rust CLI and Rust library with the same version.


reyammer (approved):
My understanding of all this is:
- the public crate names will stay as is
- since now maturin does the right thing, we don't need to rely anymore of your old publish.sh script that was implementing some hack, and so, now we can finally have the internal crate/package names (as specified by cargo.toml) to be what we want to be visible outside + we can get rid of publish.sh and just use cargo publish.

something like this?

ia0:
Yes exactly. We get rid of the last Maturin hack AFAICT.

PR #1360, opened 2026-04-17
Exit successfully when user closes stdout

### LINKED ISSUE #1354: Rust CLI panics with Broken pipe when piped to head

ia0:
Depends on #1359.

Fixes #1354. Also fixes a segfault in ONNX Runtime when shutting down while inference is running.

### DIFF
--- a/rust/cli/CHANGELOG.md
+++ b/rust/cli/CHANGELOG.md
@@ -1,9 +1,14 @@
 # Changelog
 
-## 1.0.3-dev
+## 1.1.0-dev
+
+### Minor
+
+- Exit successfully when the standard output is closed by the user
 
 ### Patch
 
+- Join all threads before shutdown to avoid segmentation faults in ONNX Runtime
 - Update dependencies
 - Fix new clippy lints
 
--- a/rust/cli/Cargo.lock
+++ b/rust/cli/Cargo.lock
@@ -440,7 +440,7 @@ dependencies = [
 
 [[package]]
 name = "magika-cli"
-version = "1.0.3-dev"
+version = "1.1.0-dev"
 dependencies = [
  "anyhow",
  "async-channel",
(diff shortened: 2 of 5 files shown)


reyammer (dismissed):

reyammer (approved):

talaumijunior-code (commented):
image

PR #1379, opened 2026-04-26
Align Python fallback client with Rust CLI (#911)

### LINKED ISSUE #911: Align the fallback python client with the new rust client
- [ ] remove compatibility mode
- [ ] when invoking the client without tty, do not use colors

Jatin-Shihora:
Fixes #911

This will remove the `--compatibility-mode` (`-c`) flag and clean up how color output is handled in the CLI.

The compatibility flag originally existed to mimic the output of the Unix `file` command. Since that behavior effectively disappeared when the `magic` field was removed in v0.6.1. Since then, the flag has only acted as a no-op alias for `--no-colors`. The Rust CLI never supported this flag, so removing it also brings both implementations closer in behavior. I tried keeping both the CLIs logical working same.

Color handling has also been updated. Colors are now automatically disabled when stdout is not a TTY, which is same as how the Rust client behaves. The `--colors` and `--no-colors` flags can still be used to explicitly override this behavior. 

Also, scripts using the Python fallback client that pass `-c` or `--compatibility-mode` will now fail with a “No such option” error.

### DIFF
--- a/python/CHANGELOG.md
+++ b/python/CHANGELOG.md
@@ -9,6 +9,10 @@ Note that for version number starting with a `0`, i.e., `0.x.y`, a bump of `x`
 should be considered as a major (and thus potentially breaking) change. See
 semver guidelines for more details about this.
 
+## [Unreleased]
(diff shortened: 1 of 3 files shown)


Jatin-Shihora:
Hi @reyammer, just a friendly follow-up on the review when you have a chance. Thanks!

PR #1415, opened 2026-06-22
rar: use application/vnd.rar instead of application/x-rar

gabriel-vasile:
`application/x-rar` is not a registered mime type. https://www.iana.org/assignments/media-types/application/vnd.rar


reyammer (approved):

ia0 (approved):

reyammer:
Merged, thank you!

PR #1419, opened 2026-07-24
Adds test for wasm

Johnchoi913:
Adds a basic test sample for wasm which while niche, did not previously exist. Generated from Rust using the Rust Playground.


talaumijunior-code (approved):

reyammer (approved):

reyammer (commented):
This is WAT (WebAssembly Text Format), not WASM (WebAssembly Binary). for magika, wasm is the binary one.

reyammer (changes requested):

reyammer:
see comment above

PR #1437, opened 2026-08-19
Replace ONNX Runtime with an embedded tract inference runtime (Rust)

ebursztein:
Replaces ONNX Runtime with [tract](https://github.com/sonos/tract) for the Rust library and CLI. The optimized NNEF model is embedded in the binary, so Rust builds no longer need to link or ship ONNX Runtime.

Python and Go continue to use the existing ONNX models and are unchanged.

### Linux end to end

Release builds of exact `origin/main` (`94ffd1d`, ONNX Runtime) and this PR were run over the same 4,000-file, 76 MiB corpus on an 8-core/16-thread Intel Xeon 2.8 GHz Google Cloud VM. Each binary used its production defaults and recursive traversal with stdout discarded, so the measurement includes startup, traversal, feature extraction, batching, inference, and ordered output.

Both binaries were warmed once. Seven measured pairs alternated which runtime ran first; the host was checked for competing build/test processes before and after each block.

| Runtime | Median wall time | Range | Median throughput |
| --- | ---: | ---: | ---: |
| `origin/main`, ONNX Runtime | 5.181 s | 5.061–5.533 s | 772 files/s |
| this PR, tract CPU | 3.730 s | 3.711–3.786 s | 1,072 files/s |

The tract CLI delivered **1.39× the throughput** and **28.0% lower wall time** end to end. The corpus construction and complete protocol are recorded in `rust/tract-bench/README.md`.

### macOS end to end

Production CLI on an M5 Max over the same 4,000-file corpus:

| Runtime | Wall time | Throughput |
| --- | ---: | ---: |
| `main`, ONNX Runtime | 1.78 s | 2,247 files/s |
| tract CPU | 1.10–1.15 s | 3,478–3,636 files/s |
| tract Metal | 0.47–0.49 s | 8,163–8,511 files/s |

The CPU path is **55–62% faster** than ONNX Runtime. Metal is **3.6–3.8× faster** end to end.

The release binary measured 22.2 MB instead of 28.1 MB. The model itself is effectively the same size: raw, gzip, and zstd-19 comparisons differ by less than 0.1%. The size reduction comes from replacing the runtime dependency, not from shrinking the model weights.

## Implementation

- Adds `magika-tract-runtime`, with a shared `Runtime` and one `Session` per resident inference thread.
- Converts ONNX to a deterministic, optimized, gzip-compressed NNEF archive for release.
- Uses fixed plans for batch classes 1, 4, 8, 16, 32, and 64, while the optimized x86_64 CPU path reuses its largest declared plan for padded tails.
- Folds the embedding activation into its constant table and specializes x86_64 LayerNorm and fused Conv1D + bias + GELU + max reduction with AVX2/FMA paths and a portable fallback.
- Lowers the GPU convolution to a fixed-shape matrix multiplication and uses tract's Metal memory arena.
- Uses one global CLI accumulator so files from all readers form full batches, with bounded ordered output.
- Selects `--backend auto|cpu|gpu`; `auto` is the default. `--threads`, `--readers`, and `--batch-size` expose the useful production controls.
- Uses all logical CPUs by default on x86_64 Linux and checked arithmetic for pipeline sizing.
- Fails closed when expected graph rewrites do not apply, scores are non-finite, or the GPU differs from the embedded CPU reference probe.

## Correctness and release gates

- Verifies all eight bundled models against ONNX Runtime at batches 1, 8, 16, 32, and 64.
- Checks CPU and GPU graph contracts and numerical parity. The largest observed conversion delta was `0.00003296137`, below the `0.0001` gate, with every argmax preserved.
- Regenerates the NNEF archive byte-for-byte and validates the embedded probe numerically on each architecture.
- Exercises the production CPU runtime on every shipped target, including an explicit AArch64 runtime-contract job, and the production GPU runtime on macOS.
- Passes stable and nightly tests, docs, formatting, and warning-denying Clippy for the modified crates; Rust 1.91 MSRV checks; release runtime tests; targeted AddressSanitizer coverage for the unsafe x86_64 packer; and x86_64/AArch64 GNU and musl cross-checks.
- Packages the runtime with the embedded models and folded embedding table, and keeps the release changelogs complete.
- Updates `event-listener` to 5.4.2 to resolve RUSTSEC-2026-0221. `cargo audit` reports no known vulnerabilities in the four Rust lockfiles; the existing transitive `paste` unmaintained warning comes through `tract-metal`.

## Compatibility and release version

This remains the decided **1.2.0** release. Reviewers should note these intentional Rust API changes from the runtime replacement:

- `Error::OrtError` is replaced by `Error::InferenceError`.
- `Session` is single-thread-owned and is no longer `Send + Sync`; callers create one session per worker.
- The deprecated `with_optimization_level` builder method remains a no-op, but no longer accepts an ONNX Runtime optimization-level type.
- Feature extraction remains async while inference is synchronous; the CLI keeps inference off async executor threads.

The benchmark utility now measures the exact shipped runtime rather than maintaining a second copy of its graph and scheduling code.


ia0 (approved):
I didn't review the tract-bench and tract-runtime crates. I think we could simplify the CLI further by having a single thread to walk, read, and batch. But I'm fine to keep the `--readers` flag just in case.

reyammer (approved):

PR #1446, opened 2026-09-07
go: free ONNX model path C string

### LINKED ISSUE #1444: Possible memory leak: model path C.CString is never freed in NewOnnx
# Possible memory leak: model path C.CString is never freed in NewOnnx

`NewOnnx` passes the model path to `CreateSession` as an inline `C.CString`. The
allocation is not bound to a variable, so there is no name available to free it
afterwards.

go/onnx/onnx_runtime.go:21

```go
	if err := C.CreateSession(ort.api, C.CString(modelPath), &ort.session, &ort.memory); err != nil {
		return nil, fmt.Errorf("create session: %v", C.GoString(C.GetErrorMessage(err)))
	}
```

go/onnx/onnx_runtime.h:15

```c
OrtStatus *CreateSession(const OrtApi *ort, const char *model, OrtSession **session, OrtMemoryInfo **memory_info) {
	...
	RETURN_ON_ERROR(ort->CreateSession(env, model, options, session));
```

`model` is taken as `const char *` and handed to `OrtApi::CreateSession`, which
copies the path. Nothing takes ownership of the buffer, and `CreateSession` does
not free it, so the string leaks on every call, on both the success and the
error path.

Fix:

```go
	cPath := C.CString(modelPath)
	defer C.free(unsafe.Pointer(cPath))
	if err := C.CreateSession(ort.api, cPath, &ort.session, &ort.memory); err != nil {
```

with `#include <stdlib.h>` in the preamble.

Separately, `RETURN_ON_ERROR` in `CreateSession` returns before releasing `env`
and `options`, and the `OrtStatus` returned to Go is never passed to
`ReleaseStatus`.

If you could credit me as a reporter for my contributions to security advisory I will be thankful.

ryux1:
## Summary

Keep the `C.CString` used for the ONNX model path in a local variable and release it after `CreateSession` returns. The deferred `C.free` covers both the successful and error paths.

This intentionally leaves the separately reported ONNX-owned environment, options, and status cleanup out of scope.

Fixes #1444

## Testing

Using Go 1.22.3 and ONNX Runtime 1.19.2 with the versions, linker flags, assets, and SHA-256 digest pinned by the repository:

- `go test ./...`
- `go vet ./...`
- `go vet -tags onnxruntime ./...`
- `go test -v -tags onnxruntime -ldflags="-linkmode=external -extldflags=-L/opt/onnxruntime/lib" ./...`
- `go build -tags onnxruntime -ldflags="-linkmode=external -extldflags=-L/opt/onnxruntime/lib" .` from `go/cli`

### DIFF
--- a/go/onnx/onnx_runtime.go
+++ b/go/onnx/onnx_runtime.go
@@ -3,11 +3,13 @@
 package onnx
 
 // #cgo LDFLAGS: -lonnxruntime
+// #include <stdlib.h>
 // #include "onnx_runtime.h"
 import "C"
 
 import (
 	"fmt"
+	"unsafe"
 )
 
 // NewOnnx returns an onnx that can perform inferences using an ONNX Runtime
@@ -18,7 +20,9 @@ func NewOnnx(modelPath string, sizeTarget int) (Onnx, error) {
 		api:        C.GetApiBase(),
 		sizeTarget: sizeTarget,
 	}
-	if err := C.CreateSession(ort.api, C.CString(modelPath), &ort.session, &ort.memory); err != nil {
+	cModelPath := C.CString(modelPath)
+	defer C.free(unsafe.Pointer(cModelPath))
+	if err := C.CreateSession(ort.api, cModelPath, &ort.session, &ort.memory); err != nil {
 		return nil, fmt.Errorf("create session: %v", C.GoString(C.GetErrorMessage(err)))
 	}
 	return ort, nil


ryux1:
Rebased onto the current `main` (`9096fd3`) and revalidated the change locally.

Passed:
- `go test ./...`
- `go vet ./...`
- `go vet -tags onnxruntime ./...`
- full `go test -v -tags onnxruntime ./...` with ONNX Runtime 1.19.2 and the standard_v3_3 assets
- tagged CLI build

PR #1448, opened 2026-09-08
test: add basic Haskell, Julia, Solidity, OCaml, and Scala samples (#662)

### LINKED ISSUE #662: Add more "basic" tests samples to cover supported content types

rksharma-owg:
## Summary
Closes #662.

Adds 5 hand-authored test samples for content types supported by Magika:
- `tests_data/basic/haskell/math_utils.hs`
- `tests_data/basic/julia/temperature_stats.jl`
- `tests_data/basic/solidity/simple_storage.sol`
- `tests_data/basic/ocaml/math_utils.ml`
- `tests_data/basic/scala/inventory.scala`

## Verification
Tested across all standard models:
- Haskell: 100% confidence
- Julia: 99.65% confidence
- Solidity: 100% confidence
- OCaml: 100% confidence
- Scala: 99.85% confidence

Python and Go test suites pass without regression.

### DIFF
--- a/tests_data/basic/haskell/math_utils.hs
+++ b/tests_data/basic/haskell/math_utils.hs
@@ -1,0 +1,23 @@
+module MathUtils where
+
+-- Factorial function with pattern matching
+factorial :: Integer -> Integer
+factorial 0 = 1
+factorial n | n > 0 = n * factorial (n - 1)
+            | otherwise = error "factorial of negative number"
+
+-- Safe division returning Maybe
+safeDiv :: Double -> Double -> Maybe Double
+safeDiv _ 0.0 = Nothing
+safeDiv x y   = Just (x / y)
+
+-- Check if a number is prime
+isPrime :: Integer -> Bool
+isPrime n
+  | n <= 1 = False
+  | otherwise = null [ x | x <- [2 .. floor (sqrt (fromIntegral n))], n `mod` x == 0 ]
+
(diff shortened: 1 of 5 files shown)


reyammer:
Thanks, but please check: https://github.com/google/magika/issues/662#issuecomment-5570674128

rksharma-owg:
Understood, thanks for the clarification. I’ll hold off on additional sample PRs while the issue is paused.

PR #1449, opened 2026-09-08
Expose the Rust library as a C library

ia0:


ia0:
Let me also add the option to identify from features.

rust/ffi/include/magika.h
```diff
@@ -0,0 +1,345 @@
...
+  /**
+   * Use GPU inference.
+   */
+  MAGIKA_BACKEND_GPU = 2,
+} MagikaBackend;
+
+/**
+ * The kind of identified file type.
+ */
+typedef enum MagikaFileTypeKind {
```
reyammer, line 59: are these reflecting existing enums in rust, or we are adding new things/abstractions? should we give more thoughts to these or it's already a done deal?

reyammer:
Left a quick comment. do you want me to review something specific, or just need a stamp?

rust/ffi/include/magika.h
ia0: They reflect existing enums in Rust, but enums in Rust may have data: https://docs.rs/magika/latest/magika/enum.FileType.html

So in C, those enums are just "tags" that are used in structs that reflect the Rust enums.

The idea is that the C API follows as closely as possible the Rust API, except that it tries before all to look like an idiomatic C API.

ia0:
> Left a quick comment. do you want me to review something specific, or just need a stamp?

No, I don't think there's anything specific to review. This is a "copy/paste" of the Rust API as an idiomatic C API.

However, I'd like to merge #1461 first. I'll send it you way once it passes the CI. It fixes the issue we observed about stuck workflows, introduced by #1437.

reyammer:
ack, makes sense. thanks!

reyammer (dismissed):

reyammer (approved):

PR #1447, opened 2026-09-07
Prepare Magika 2.0 with opt-in rules, deferred runtimes and reproducible benchmarks

ebursztein:
Prepare the Rust CLI and library as **Magika 2.0 (`2.0.0-dev`)** with the breaking Runtime/Session API, optional format rules and deferred inference libraries.

Rules inspect at most 4 KiB and require minimum observed prefixes plus format-specific structural checks. `--rules=enforce` falls through to ML on misses or conflicts; `--rules=only` abstains and skips inference initialization. Unqualified rules remain disabled. The [rule reference](https://github.com/google/magika/blob/worktree/rules-pr/rules/QUALITY.md) documents scope and sources.

Auto and GPU CLI modes process on CPU while GPU prepares asynchronously. Once ready, GPU mode prefers GPU; Auto admits batches of at least eight files. Explicit CPU avoids GPU loading. Standard CLI archives/installers and binary Python wheels include their inference libraries. Native Windows rules and Windows CUDA qualification holds remain documented.

The benchmark records tool/version/config, named datasets, timestamps, raw predictions and whole-process Hyperfine timings using shipping resource defaults. Compressed measurements are canonical; reports regenerate from JSON. TrID has separate verified configurations with and without StringZilla. Earlier mislabeled TrID rows are corrected from their runtime evidence.

- [Benchmark catalogue and reproduction](https://github.com/google/magika/blob/worktree/rules-pr/rules/benchmarks/README.md)
- [Combined corpus table](https://github.com/google/magika/blob/worktree/rules-pr/rules/benchmarks/reports/2026-09-09-65505ce3-cpu-handoff-combined/overview.md) · [Sembiance v3 table](https://github.com/google/magika/blob/worktree/rules-pr/rules/benchmarks/reports/2026-09-09-65505ce3-cpu-handoff-sembiance/overview.md)
- [Combined TrID comparison](https://github.com/google/magika/blob/worktree/rules-pr/rules/benchmarks/reports/2026-09-09-7599ca96-trid-combined/trid-comparison.md) · [Sembiance TrID comparison](https://github.com/google/magika/blob/worktree/rules-pr/rules/benchmarks/reports/2026-09-09-7599ca96-trid-sembiance/trid-comparison.md)
- [Preserved handoff and historical engineering evidence](https://github.com/google/magika/blob/worktree/rules-pr/rules/benchmarks/handoffs/README.md)

[...]

PR #1464, opened 2026-09-13
Harden the CLI against bad limits, special files and pipeline errors

ebursztein:
Part of the split of #1447 (plan in #1462, Part A row 2). Three CLI robustness fixes, one commit each, ported by hand from #1447 (`c4762f68`, `e8b99770`) because the CLI has moved since.

## Fixes

1. **Resource limits are bounded when parsing.** `--batch-size`, `--threads` and `--readers` size channels and spawn threads, and only zero was rejected: `--batch-size 18446744073709551615` panicked (exit 101). They are now 1–64 and 1–256, and a bad value is a usage error (exit 2) naming the flag.
2. **Special files are rejected before a reader sees them.** A named pipe found during `-r` blocked its reader forever and stalled the run. Pipes, sockets and devices are now reported as `unsupported_file_type`, and the files around them are still classified. Nothing that worked stops working: readers need random access, so on main even `magika <(cat file)` already failed with `unknown`.
3. **Output and failures survive pipeline errors.** A pipeline error dropped results still being reordered and left a JSON array unclosed; a failed input called `process::exit`, skipping thread joins; a non-UTF-8 path panicked building JSON; a wrong inference row count was only a `debug_assert`. Printing now keeps received results, closes the array and returns the first failure, and a later broken pipe no longer hides it. **Visible change:** a run with failed inputs now ends with `Error: one or more input files failed` on stderr (exit code still 1).

## Tests

- `rust/cli/tests/resource_limits.rs`: fails on main (panic), passes here.
- `rust/cli/tests/special_files.rs`: a pipe inside a directory and a pipe, symlinked pipe and socket passed directly; both hang on main, pass here.
- Six unit tests of printing and dispatch in `main.rs`.
- `rust/cli/test.sh` now runs `cargo test` (same commit as in #1463) and is green locally.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ia0 (dismissed):

ia0:
I've restricted the PR scope to the non-regular file fix, the JSON UTF-8 file name fix, and the hardening of the experimental flags. I've removed the added complexity about pipeline errors. Those should never happen (which means that if they do, the program is expected to panic, and all output should be ignored). In particular, we should not try to "fix the output" (whatever that could mean). Note that this is very different from file errors. Those are expected to happen and have a defined representation in the output (in particular the program does not panic, does not write to stderr, but does return a non-zero error code to raise the user attention).

I've also fixed a corner case where file errors could hide pipeline errors after everything has been printed.

ia0 (approved):

ia0-review (approved):

PR #1471, opened 2026-09-13
Add 109 content types for format rules

ebursztein:
Part of the split of #1447 (tracker: https://github.com/google/magika/pull/1447#issuecomment-5656016366). Needed by the rules integration on top of #1462, and independent of it.

## What

Content types that format rules can name, without changing the model:

| | New in the Rust library |
|---|---:|
| New knowledge base entries (`assets/content_types_kb.min.json`) | 50 |
| Existing knowledge base entries now exposed (`rust/gen/content_types`) | 47 |
| Rule-only outputs the knowledge base does not describe (`rust/gen/rule-content-types.json`) | 12 |
| **Total** | **109** |

Python and JavaScript expose the whole knowledge base, so they gain its 50 new labels.

## Commits

1. **Generate content types that only rules name.** `rust/gen` reads `rule-content-types.json` (moved from #1447's `rules/content-types.json`), requires each entry to be selected and absent from the knowledge base, and requires every selected label to resolve. The generated `ContentType` gains `from_label()`, and underscored labels get camel-case variants (`postgres_dump` → `PostgresDump`). Ported onto the generator as restructured by #1449; with the file empty, only `from_label()` is generated.
2. **Add 109 content types.** Only the inputs are taken from #1447; every output is regenerated with `rust/gen` and `python/scripts/sync.py`.

## Proof

- `rust/lib/src/content.rs`, `rust/lib/src/model.rs`, `python/src/magika/types/content_type_label.py`, `js/src/content-type-label.ts` and `js/src/content-types-infos.ts` are byte-identical to #1447's. `rust/ffi/src/content.rs` is new since #1447 and regenerates with the new labels.
- Every existing label keeps its threshold and overwrite target; each new label has threshold 0.5 and maps to itself, so model outputs are unchanged.
- Re-running the generator leaves the tree clean.
- `rust/lib/test.sh`, `rust/gen/test.sh`, ffi clippy, and the Python module and reference-inference tests: green locally. The JS package was not built locally.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rust/gen/src/main.rs
```diff
@@ -49,6 +49,17 @@ fn filter_content_types(
     let content_types_content = std::fs::read_to_string("content_types")?;
     let mut labels = content_types_content.lines().collect::<BTreeSet<_>>();
     labels.extend(model_config.target_labels_space.iter().map(|x| x.as_str()));
+    // Outputs that only rules name and the knowledge base does not describe.
+    let rule_only: BTreeMap<String, ContentType> =
+        serde_json::from_reader(File::open("rule-content-types.json")?)?;
```
ia0, line 54: Maybe a naive question but why not have those in the knowledge base? Is the KB only for machine learning?

We would anyway need magika-rules to provide a list of labels it may produce. But that's probably going to be for the PR that starts using magika-rules (which I don't think there is yet?).

So we probably don't need to touch anything in Rust in this PR. That would be in the PR that starts using magika-rules.

rust/gen/src/main.rs
```diff
@@ -100,6 +111,19 @@ fn generate_lib_content(
     writeln!(output, "}}\n")?;
     writeln!(output, "impl ContentType {{")?;
     writeln!(output, "    pub(crate) const SIZE: usize = {};\n", variants.len())?;
+    writeln!(output, "    /// Looks up an exact, canonical content label for a regular file.")?;
```
ia0, line 114: This part of the PR can be its own PR.

ia0 (changes requested):

ia0 (approved):
I should not be reviewer since there's no Rust anymore. But it still looks good to me.

rust/gen/src/main.rs
ebursztein: Agreed, no reason to keep them apart: the knowledge base describes content types whatever identifies them. The 12 rule-only types are now in it, and this PR no longer touches Rust. The rules integration will expose in Rust the labels magika-rules can produce.

rust/gen/src/main.rs
ebursztein: Done, moved to #1484 (now merged).

reyammer (approved):

